Issue #67, opened 2016-11-23
Move release notes to this repo


luxas:
Right now they are in https://github.com/kubernetes/kubernetes.github.io/blob/master/docs/admin/kubeadm.md#releases-and-release-notes, which is a bad place.

Now when we have this repo, they should be moved over here.

mikedanese:
I disagree with this. Docs should remain in the docs repo so they are published to the site.

luxas:
They should never have gone into the website docs, we just had no other place to put them at the time.
Why wouldn't we move them over here in a CHANGELOG like the main repo?
It's not an user-facing document, but more a developer facing doc.

mikedanese:
Release notes are definitely user facing. Release notes are meant to be read by users before upgrading. It would be very nice to publish them on the site.

luxas:
I just think we should be consistent with all other projects.
The main kubernetes repo and minikube both have the CHANGELOG in the repo, not in the docs

We would of course link to the changelog/release notes from the site, but they should not be there IMO.

If we want all changelogs in the docs repo, we should be consistent and do it for all projects in that case.

luxas:
Fixed with #81

PR #81, opened 2016-12-02
Move the CHANGELOG over here like all the other k8s repos

luxas:
ref: #67


luxas:
@mikedanese If you could merge this now, I'd be grateful and able to move forward on some things.

PR #100, opened 2017-01-02
WIP: Initial kubeadm api proposal

luxas:
This is far from done, but posting progress on my work on the kubeadm API here.

This is kind of a kubeadm-to-beta design doc which we have to discuss in detail.
I'll improve it as we go.

cc @jbeda @mikedanese @dgoodwin @errordeveloper @lukemarsden @dgoodwin @kad @roberthbailey


docs/design/kubeadm-api.md
```diff
@@ -0,0 +1,275 @@
...
+In this process, I've been thinking a lot about security and tried to put that first. Also, I've read the 
+kops, kargo and kube-up code in detail to see how they are doing these things. We don't want to make an API that's incompatible with those tools.
+
+So far (for kubeadm) I've identified these:
+
+#### Certificates
+
+First, certificates should be created in a directory.
+There should be:
+ - a CA certificate (`ca.pem`) with its private key (`ca-key.pem`)
```
mikedanese: nit: we should try to name these `ca.key` and `ca.crt`. While they are pem encoded, these are the more common names for this type of file.

docs/design/kubeadm-api.md
```diff
@@ -0,0 +1,275 @@
...
+   - Is the public key necessary to have here (`ca-pub.pem`)?
```
mikedanese: ca.crt contains the public key.

docs/design/kubeadm-api.md
```diff
@@ -0,0 +1,275 @@
...
+ - an API Server certificate (`apiserver.pem`) signed with `ca.pem` as the CA with its private key (`apiserver-key.pem`)
+ - a Service Account private key that's used for signing Service Accounts' tokens.
```
mikedanese: We've used the apiserver.key to do this so far.

docs/design/kubeadm-api.md
```diff
@@ -0,0 +1,275 @@
...
+   - TODO: Is the certificate for the serviceaccount private key necessary?
+
+Also, we should make sure that both RSA and ECDSA private keys work with kubeadm and have tests for both.
+We probably want to expose the private key bitsize also.
```
mikedanese: How and why?

docs/design/kubeadm-api.md
```diff
@@ -0,0 +1,275 @@
...
+
+One more option that could be useful to many would be to let the user specify their own ca.pem and ca-key.pem,
```
mikedanese: Hmm, I thought this was already possible. This is absolutely a requirement.

mikedanese: If these exist in the location that kubeadm would create them, then kubeadm should just use them. No need to copy. This should be the general pattern for how we implement phases w.r.t. reentrance

docs/design/kubeadm-api.md
```diff
@@ -0,0 +1,275 @@
+## Kubeadm Phases and API
+
+### Refactoring kubeadm
+We all want kubeadm modular and componentized, here’s the proposal for that!
```
roberthbailey: s/,/;

docs/design/kubeadm-api.md
```diff
@@ -0,0 +1,275 @@
...
+   - `sa-key.pem`
+
+
+#### Generate KubeConfig files for the admin and the kubelet on the master
+
+The second phase takes the certificates generated in phase 1 and the public master endpoint as input and produces two kubeconfig files,
+`admin.conf` and `kubelet.conf` in `/etc/kubernetes/` by default.
+
+A general question here, do we want the KubernetesDir to be configurable at all or does it only add complexity/confusion if a user ever changed that path?
+I lean towards having that as a constant, since that seems to be the case with a lot of other deployments and the gain is very small.
```
mikedanese: I also lean towards leaving this unconfigurable. If there is an uproar we can always make it configurable later.

docs/design/kubeadm-api.md
```diff
@@ -0,0 +1,275 @@
+## Kubeadm Phases and API
+
+### Refactoring kubeadm
+We all want kubeadm modular and componentized, here’s the proposal for that!
+I’m not writing down all reasons for doing this; they are already available in the blog post: https://docs.google.com/document/d/1bpEGhdg8t-2Ae-IL_ezZIPklMHpZLWwnoAPMgLwT3oU
```
roberthbailey: linkify blog post rather than putting the https after the :

docs/design/kubeadm-api.md
```diff
@@ -0,0 +1,275 @@
...
+type MasterConfiguration struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// Data shared between phases
+	Master     Master     `json:"master"`
+	Networking Networking `json:"networking"`
+
+	// Phases
+	Certificates   Certificates   `json:"certificates"`
+	MakeKubeConfig MakeKubeConfig `json:"makeKubeConfig"`
```
mikedanese: Drop the `Make` verb. The naming of the phases should be consistent.

docs/design/kubeadm-api.md
```diff
@@ -0,0 +1,275 @@
...
+
+### Phases
+We want to identify the phases every single Kubernetes deployment goes through in order to get the cluster up and running.
+
+In this process, I've been thinking a lot about security and tried to put that first. Also, I've read the 
+kops, kargo and kube-up code in detail to see how they are doing these things. We don't want to make an API that's incompatible with those tools.
+
+So far (for kubeadm) I've identified these:
+
+#### Certificates
```
roberthbailey: @cjcullen 

We should also create certificates for master -> kubelet connections, which for kube-up we are doing with a second "CA universe".

docs/design/kubeadm-api.md
roberthbailey: If we set the bitsize to be large enough, we shouldn't need to expose other options (it's just more confusing/complexity for users). If you really want a different bitsize, you should follow the "bring your own cert" path instead of letting us generate certs, as then you have complete control over the algorithm and the bitsize.

docs/design/kubeadm-api.md
```diff
@@ -0,0 +1,275 @@
...
+
+Also, we should make sure that both RSA and ECDSA private keys work with kubeadm and have tests for both.
+We probably want to expose the private key bitsize also.
+
+One more option that could be useful to many would be to let the user specify their own ca.pem and ca-key.pem, 
+and kubeadm would just copy them over to the destination directory and proceed as usual. This might be beneficial to orgs
+that require a specific CA but lets the apiserver and serviceaccount keys/certs be generated by kubeadm.
+
+**Inputs**
+ - `AdvertiseAddresses` is needed for knowing which IPs the certs should be signed for
```
roberthbailey: Why are these inputs separated this way? They are all specifying "names which should be signed" and should be possible to group into a single input (and let the code determine if they are ip addresses or dns names and do the right thing).

docs/design/kubeadm-api.md
```diff
@@ -0,0 +1,275 @@
...
+#### Generate KubeConfig files for the admin and the kubelet on the master
+
+The second phase takes the certificates generated in phase 1 and the public master endpoint as input and produces two kubeconfig files,
+`admin.conf` and `kubelet.conf` in `/etc/kubernetes/` by default.
+
+A general question here, do we want the KubernetesDir to be configurable at all or does it only add complexity/confusion if a user ever changed that path?
+I lean towards having that as a constant, since that seems to be the case with a lot of other deployments and the gain is very small.
+
+There also is the question whether this phase should generate both the kubelet and the admin config. Or should those be two different phases/steps?
+How can we make the kubelet have just a little permissions but still use it for bootstrapping the api server? Can start the master ("bootstrap") kubelet with full access,
```
roberthbailey: Does bootkube do anything here? I agree that we want the kubelet to end up with "dropped permissions" even if it has to start out with root permissions to get things up and running.

luxas: Nope, bootkube has only one kubeconfig file for everything as I interpreted the code

docs/design/kubeadm-api.md
luxas: This was just the way the arguments are exposed now, my intention was indeed to have just a common field for both DNS and IPs. Suggestions on the name?

docs/design/kubeadm-api.md
luxas: It should really soon be up on the site so waiting a little bit before updating that, it was never meant to be the doc, but in absence of the real blog post, I temporarily linked to the doc

docs/design/kubeadm-api.md
luxas: fixed

docs/design/kubeadm-api.md
luxas: fixed, let's discuss the second sentence more generally

docs/design/kubeadm-api.md
luxas: Cool, then we agree :+1:

docs/design/kubeadm-api.md
luxas: It's not possible right now (like other things that are half-implemented). Updated the statement.

docs/design/kubeadm-api.md
luxas: Thanks, I've been trying to find the balance all the time for what should and should not be exposed.
While reading your arguments, it's clear that this falls under the bring-your-own certs thing.
I appreciate that feedback extra much

docs/design/kubeadm-api.md
luxas: What's the difference between having a dedicated private key or reusing the apiserver's?
Is there any difference in security?

docs/design/kubeadm-api.md
luxas: Yes, that was what I thought. I just couldn't understand why we had both `ca.pem` and `ca-pub.pem`, I'm gonna remove the public key only one because it's unnecessary

docs/design/kubeadm-api.md
luxas: Where is the code for this? Is using a separate CA required?
I'd very much like to have more information about how you have it setup on GCE regarding that.

luxas:
Thanks @roberthbailey and @mikedanese!

Answered your comments and updated a somehow updated version.
Much still to write down for me however.

I appreciate the comments; let's do this together and come up with a rock-solid API!

docs/design/kubeadm-api.go
```diff
@@ -0,0 +1,211 @@
...
+type MasterConfiguration struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// Data shared between phases. Used for defaulting
+	Master     Master     `json:"master"`
+	Networking Networking `json:"networking"`
+	CertificatesDir string `json:"certificatesDir"`
+
+	// Phases
+	Certificates     Certificates     `json:"certificates"`
```
pires: `Certificates` is not clear about its intents. If we have certificates then why do we need `CertificatesDir`? And while I may understand the design, the API itself doesn't seem self-explanatory and I believe will create confusion.

docs/design/kubeadm-api.go
```diff
@@ -0,0 +1,211 @@
...
+	Addons           Addons           `json:"addons"`
+}
+
+type Phase struct {
+	Annotations map[string]string `json:"annotations"`
+}
+
+// Shared
+// TODO: This field should be HA friendly
+type Master struct {
```
pires: `MasterParameters` or similar?

docs/design/kubeadm-api.go
```diff
@@ -0,0 +1,211 @@
+package main
+
+type MasterConfiguration struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// Data shared between phases. Used for defaulting
+	Master     Master     `json:"master"`
```
luxas: This is not by far done; I'm trying to see how I can move stuff around to get rid of it

luxas: Leftover from the current api

docs/design/kubeadm-api.go
luxas: We might group all phases under a Phase(s) struct to make things more clear.
Here, `Certificates` stands for the first phase whose task is to create certs.
`CertificatesDir` is a param to the `Certificates` phase that says where certs should be put

docs/design/kubeadm-api.go
luxas: Yes, this should change indeed.

mikedanese: MasterConfiguration is the convention we use for other componentconfig structs.

luxas: yes, that stays the same.

This Master struct that's here was named API before, but should not exist at all I think

mikedanese:
why can't we pass the full config into every phase, e.g.:
```
$ cat cfg.json
masterConfiguration:
  phaseA:
    foo: "abc"
    **bar: "def"**
  phaseB:
    **bar: "def"**
    baz: "ghi"
$ kubeadm phase a --config cfg.json
$ kubeadm phase b --config cfg.json
```
Later phases should be able to consume configs from earlier phases, including defaults and randomly generated secrets. This might mean writing to the config per phase.

docs/design/kubeadm-api.go
```diff
@@ -0,0 +1,251 @@
...
+	// Phases
+	Certificates     Certificates     `json:"certificates"`
+	KubeConfig       KubeConfig       `json:"kubeConfig"`
+	ControlPlane     ControlPlane     `json:"controlPlane"`
+	Discovery        MasterDiscovery        `json:"discovery"`
+	APIConfiguration APIConfiguration `json:"apiConfiguration"`
+	Addons           Addons           `json:"addons"`
+}
+
+type Phase struct {
```
mikedanese: PhaseMetadata?

luxas:
Closing in favor for: #156

PR #156, opened 2017-02-09
Add a doc about the kubeadm design and phases implementation

luxas:
This document describes the functions of kubeadm and how the phases should be splitted up

@lukemarsden @roberthbailey @mikedanese @errordeveloper @pires @dmmcquay @jbeda

### DIFF
--- a/docs/design/design.md
+++ b/docs/design/design.md
@@ -1,0 +1,261 @@
+## Implementation design for kubeadm
+
+`kubeadm init` and `kubeadm join` together provides a nice user experience for creating a best-practice but bare Kubernetes cluster from scratch.
+However, it might not be obvious _how_ kubeadm does that.
+
+This document strives to explain the phases of work that happen under the hood.
+Also included is ComponentConfiguration API types for talking to kubeadm programmatically.
+
+**Note:** Each and every one of the phases must be idempotent!
+
+### The scope of kubeadm
+
+The scope of `kubeadm init` and `kubeadm join` is to provide a smooth user experience for the user while bootstrapping a best-practice cluster.
+
+The cluster that `kubeadm init` and `kubeadm join` set up should be:
+ - Secure
+   - It should adopt latest best-practices like
+     - enforcing RBAC
+     - using the Node Authorizer
+     - using secure communication between the control plane components
+     - using secure communication between the API Server and the kubelets
+     - making it possible to lock-down the kubelet API
+     - locking down access to the API system components like the kube-proxy and kube-dns
+     - locking down what a Bootstrap Token can access
+ - Easy to use
+   - The user should not have to run anything more than a couple of commands, including:
+     - `kubeadm init` on the master
+     - `export KUBECONFIG=/etc/kubernetes/admin.conf`
+     - `kubectl apply -f <network-of-choice.yaml>`
+     - `kubeadm join --token <token> <master>`
+     - The `kubeadm join` request to add a node should be automatically approved
+ - Extendable
+   - It should for example _not_ favor any network provider, instead configuring a network is out-of-scope
+   - Should provide a config file that can be used for customizing various parameters
+
+#### A note on constants / well-known values and paths
+
+We have to draw the line somewhere about what should be configurable, what shouldn't, and what should be hard-coded in the binary.
+
+We've decided to make the Kubernetes directory `/etc/kubernetes` a constant in the application, since it is clearly the given path in a majority of cases,
+and the most intuitive location. Having that path configurable would confuse readers of an on-top-of-kubeadm-implemented deployment solution.
+
+This means we aim to standardize:
+ - `/etc/kubernetes/manifests` as the path where kubelet should look for Static Pod manifests
+   - Temporarily when bootstrapping, these manifests are present:
+     - `etcd.yaml`
+     - `kube-apiserver.yaml`
+     - `kube-controller-manager.yaml`
+     - `kube-scheduler.yaml`
+ - `/etc/kubernetes/kubelet.conf` as the path where the kubelet should store its credentials to the API server.
+ - `/etc/kubernetes/kubelet.conf` as the path from where the admin can fetch his/her superuser credentials.
+
+
+## `kubeadm init` phases
+
+### Phase 1: Generate the necessary certificates
+
+`kubeadm` generates certificate and private key pairs for different purposes.
+Certificates are stored by default in `/etc/kubernetes/pki`. This directory is configurable.
+
+There should be:
+ - a CA certificate (`ca.crt`) with its private key (`ca.key`)
+ - an API Server certificate (`apiserver.crt`) using `ca.crt` as the CA with its private key (`apiserver.key`). The certificate should:
+   - be a serving server certificate (`x509.ExtKeyUsageServerAuth`)
+   - contain altnames for
+     - the kubernetes' service's internal clusterIP and dns name (e.g. `10.96.0.1`, `kubernetes.default.svc.cluster.local`, `kubernetes.default.svc`, `kubernetes.default`, `kubernetes`)
+     - the hostname of the node
+       - **TODO:** I guess this might be a requested feature in opinionated setups, but might be a no-no in more advanced setups. Consensus here?
+     - the IPv4 address of the default route
+     - optional extra altnames that can be specified by the user
+ - a client certificate for the apiservers to connect to the kubelets securely (`apiserver-kubelet-client.crt`) using `ca.crt` as the CA with its private key (`apiserver-kubelet-client.key`). The certificate should:
+   - be a client certificate (`x509.ExtKeyUsageClientAuth`)
+   - be in the `system:masters` organization
+ - a private key for signing ServiceAccount Tokens (`sa.key`) along with its public key (`sa.pub`)
+ - a CA for the front proxy (`front-proxy-ca.crt`) with its key (`front-proxy-ca.key`)
+ - a client cert for the front proxy client (`front-proxy-client.crt`) using `front-proxy-ca.crt` as the CA with its key (`front-proxy-client.key`)
+
+
+If a given certificate and private key pair both exist, the generation step will be skipped and those files will be validated and used for the prescribed use-case.
+This means the user can, for example, prepopulate `/etc/kubernetes/pki/ca.{crt,key}` with an existing CA, which then will be used for signing the rest of the certs.
+
+### Phase 2: Generate KubeConfig files for the master components
+
+There should be:
+ - a KubeConfig file for kubeadm to use itself and the admin: `/etc/kubernetes/admin.conf`
+   - the "admin" here is defined as `kubeadm` itself and the actual person(s) that is administering the cluster and want to control the cluster
+     - with this file, the admin has full control (**root**) over the cluster
+   - inside this file, a client certificate is generated from the `ca.crt` and `ca.key`. The client cert should:
+     - be a client certificate (`x509.ExtKeyUsageClientAuth`)
+     - be in the `system:masters` organization
+     - include a CN, but that can be anything. `kubeadm` uses the `kubernetes-admin` CN.
+ - a KubeConfig file for kubelet to use: `/etc/kubernetes/kubelet.conf`
+   - inside this file, a client certificate is generated from the `ca.crt` and `ca.key`. The client cert should:
+     - be a client certificate (`x509.ExtKeyUsageClientAuth`)
+     - be in the `system:nodes` organization
+     - have the CN `system:node:<hostname-lowercased>`
+ - a KubeConfig file for controller-manager: `/etc/kubernetes/controller-manager.conf`
+   - inside this file, a client certificate is generated from the `ca.crt` and `ca.key`. The client cert should:
+     - be a client certificate (`x509.ExtKeyUsageClientAuth`)
+     - have the CN `system:kube-controller-manager`
+ - a KubeConfig file for scheduler: `/etc/kubernetes/scheduler.conf`
+   - inside this file, a client certificate is generated from the `ca.crt` and `ca.key`. The client cert should:
+     - be a client certificate (`x509.ExtKeyUsageClientAuth`)
+     - have the CN `system:kube-scheduler`
+
+`ca.crt` is also embedded in all the KubeConfig files.
+
+### Phase 3: Bootstrap the control plane by using Static Pods
+
+#### etcd
+
+Determine if the user specified external etcd options. If not, etcd should:
+ - be spun up as a Static Pod
+ - listen on `localhost:2379` and use `HostNetwork=true`
+ - have `PodSpec.SecurityContext.SELinuxOptions.Type=spc_t` set because of https://github.com/kubernetes/kubeadm/issues/107
+ - be at a minimum version `3.0.14`
+ - make a `hostPath` mount out from the `dataDir` to the host's filesystem
+
+#### API Server
+
+The API Server needs to know this in particular:
+ - The subnet to use for services
+ - Where to find the etcd server
+ - The address and port to bind to; defaults to the IP Address of the default interface and port 6443 for secure communication
+ - Any extra flags and/or HostPath Volumes/VolumeMounts specified by the user
+
+Other flags that are set:
+ - The `BootstrapTokenAuthenticator` authentication module is enabled
+ - `--client-ca-file` to `ca.crt`
+ - `--tls-cert-file` to `apiserver.crt`
+ - `--tls-private-key-file` to `apiserver.key`
+ - `--kubelet-client-certificate` to `apiserver-kubelet-client.crt`
+ - `--kubelet-client-key` to `apiserver-kubelet-client.key`
+ - `--service-account-key-file` to `sa.pub`
+ - `--requestheader-client-ca-file` to `front-proxy-ca.crt`
+ - `--admission-control` to `NamespaceLifecycle,LimitRanger,ServiceAccount,PersistentVolumeLabel,DefaultStorageClass,ResourceQuota`
+   - ...or whatever the recommended set of admission controllers is at a given version
+ - `--storage-backend` to `etcd3`. Support for `etcd2` in kubeadm is dropped.
+ - `--kubelet-preferred-address-types` to `InternalIP,ExternalIP,Hostname`
+   - This makes `kubectl logs` and other apiserver -> kubelet communication work in environments where the hostnames of the nodes aren't resolvable
+ - `--requestheader-username-headers=X-Remote-User`, `--requestheader-group-headers=X-Remote-Group`, `--requestheader-extra-headers-prefix=X-Remote-Extra-`, --requestheader-allowed-names=front-proxy-client` so the front proxy (API Aggregation) communication is secure.
+
+
+#### Controller Manager
+
+The controller-manager needs to know this in particular:
+ - The Pod Network CIDR if any; also enables the Subnet Manager feature (required for some CNI network plugins)
+
+Other flags that are set unconditionally:
+ - The `BootstrapSigner` and `TokenCleaner` controllers are enabled
+ - `--root-ca-file` to `ca.crt`
+ - `--cluster-signing-cert-file` to `ca.crt`
+ - `--cluster-signing-key-file` to `ca.key`
+ - `--service-account-private-key-file` to `sa.key`
+ - `--use-service-account-credentials` to `true`
+
+#### Scheduler
+
+kubeadm doesn't set any special scheduler flags.
+
+Common properties for the control plane components:
+ - Leader election is enabled for both the controller-manager and the scheduler
+ - `HostNetwork: true` is present on all static pods since there is no network configured yet
+
+#### Wait for the control plane to come up
+
+This is a critical moment in time for kubeadm clusters.
+kubeadm waits until `localhost:6443/healthz` returns `ok`
+
+kubeadm relies on the kubelet to pulĺ the control plane images and run them properly as Static Pods.
+But there are (as we've seen) a lot of things that can go wrong. Most of them are network/resolv.conf/proxy related.
+
+### Phase 4+: Post-bootstrap Phases
+
+kubeadm completes a couple of tasks also after the control plane is up, namely these tasks:
+
+#### Marks where the master is
+
+This addon essentially does just this in pseudo-code:
+
+```bash
+kubectl taint node ${master_name} node-role.kubernetes.io/master:NoSchedule
+kubectl label node ${master_name} node-role.kubernetes.io/master=""
+```
+
+#### cluster-info
+
+This phase creates the `cluster-info` ConfigMap in the `kube-public` namespace as defined in [the Bootstrap Tokens proposal](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/bootstrap-discovery.md)
+ - The `ca.crt` and the address/port of the apiserver is added to the `cluster-info` ConfigMap in the `kubeconfig` key
+ - Exposes the `cluster-info` ConfigMap to unauthenticated users (i.e. users in RBAC group `system:unauthenticated`)
+
+**Note:** The access to the `cluster-info` ConfigMap _is not_ rate-limited. 
+This may or may not be a problem if you expose your master to the internet.
+Worst-case scenario here is a DoS attack where an attacker uses all the in-flight requests the kube-apiserver can handle to serving the `cluster-info` ConfigMap.
+TBD for v1.8
+
+#### self-hosting
+
+Parses the yaml static pod manifests in `/etc/kubernetes/manifests` and converts them to DaemonSets with master affinity:
+ - TODO
+
+#### kube-proxy
+
+A ServiceAccount for `kube-proxy` is created in the `kube-system` namespace.
+
+Deploy kube-proxy as a DaemonSet:
+ - the credentials (`ca.crt` and `token`) to the master come from the ServiceAccount
+ - the location of the master comes from a ConfigMap
+ - the `kube-proxy` ServiceAccount is bound to the privileges in the `system:node-proxier` ClusterRole
+
+#### kube-dns
+
+A ServiceAccount for `kube-dns` is created in the `kube-system` namespace.
+
+Deploy the kube-dns Deployment and Service:
+ - it's the upstream kube-dns deployment relatively unmodified
+ - the `kube-dns` ServiceAccount is bound to the privileges in the `system:kube-dns` ClusterRole
+
+#### tls-bootstrap
+
+The TLS Bootstrap ClusterRole (`system:node-bootstrapper`) is bound to the `system:bootstrappers` Group so Bootstrap Tokens are able to access the CSR API.
+
+The `system:bootstrappers` Group is granted auto-approving status by it being able to `POST /apis/certificates.k8s.io/certificatesigningrequests/nodeclient`.
+ - The auto-approving certificate controller in the controller-manager checks whether the poster of the CSR (in this case the Bootstrap Token) can POST to
+   `/apis/certificates.k8s.io/certificatesigningrequests/nodeclient`. If the poster can, the controller approves the CSR.
+ - This makes it possible to easily revoke the auto-approving functionality by removing the `ClusterRoleBinding` that grants Bootstrap Tokens that, or you can
+   revoke access for all Bootstrap Tokens and instead make the auto-approving more granular by granting just a few users or tokens access to auto-approved credentials. 
+
+## `kubeadm join` phases
+
+### Phase 1: Fetch the `cluster-info` ConfigMap
+
+This phase is skipped if
+a) the `cluster-info` ConfigMap isn't exposed publicly (or created at all)
+b) the user specified a file with the required `cluster-info` information
+
+In the future, we can omit skipping the b) case and in the case valid information already is passed, refresh that information about the cluster location once again at
+join time.
+
+This phase basically issues a `GET /api/v1/namespaces/kube-public/configmaps/cluster-info` and validates the information there by using the token. You can find more
+details about exactly how it does that in the Bootstrap Token discovery proposal.
+
+## Phase 2: Do the TLS Bootstrap flow
+
+`kubeadm` posts a CSR to the API server, which is then approved and signed by the certificates controller in the controller-manager.
+`kubeadm` then writes the signed client certificate credentials to `/etc/kubernetes/kubelet.conf` for consumption by the kubelet.
+
+**TODO:** It is planned that the kubelet will handle this phase 2 on its own in v1.8
+
+## Extending `kubeadm`
+
+There are a two primary ways to extend `kubeadm`:
+ - By setting CLI arguments or editing the lightweight `kubeadm init` API.
+ - By running the phases you need separately and giving every phase the arguments it needs
+
+The `kubeadm init` and `kubeadm join` APIs respectively are very limited in scope.
+They are there to make it possible to customize a couple of things to your needs, but doesn't allow for full flexibility.
+
+### Open Questions
+
+What do we have to change in this proposal/design doc to make kubeadm HA-friendly?


docs/design/phases.md
```diff
@@ -0,0 +1,110 @@
+## Phases in kubeadm
+
+`kubeadm init` does a lot of things. In order to make kubeadm the common building block for Kubernetes we want it to be, each logical phase kubeadm does during `init` should be invokable separately. Here's an explanation what kubeadm does in the different phases:
```
dmmcquay: Suggested edit:

`kubeadm init` does a lot. In order to make kubeadm the common building block for Kubernetes, each logical phase kubeadm does during `init` should be invokable separately. Here's an explanation what kubeadm does in these different phases:

docs/design/phases.md
```diff
@@ -0,0 +1,110 @@
...
+
+#### A note on constants / well-known values and paths
+
+Somewhere we have to draw the line what should be configurable, what shouldn't and what should be hard-coded in the binary.
```
dmmcquay: Suggested edit:

We have to draw the line somewhere about what should be configurable, what shouldn't, and what should be hard-coded in the binary.

docs/design/phases.md
```diff
@@ -0,0 +1,110 @@
...
+
+We've decided to make the Kubernetes directory `/etc/kubernetes` a constant in the application, since it is clearly the given path in a lot of places,
+and the most intuitive location. Having that path configurable would for sure confuse readers of a on-top-of-kubeadm-implemented deployment solution.
```
dmmcquay: Suggested edit:

and the most intuitive location. Having that path configurable would confuse readers of an on-top-of-kubeadm-implemented deployment solution.

docs/design/phases.md
```diff
@@ -0,0 +1,110 @@
+## Phases in kubeadm
+
+`kubeadm init` does a lot of things. In order to make kubeadm the common building block for Kubernetes we want it to be, each logical phase kubeadm does during `init` should be invokable separately. Here's an explanation what kubeadm does in the different phases:
+
+#### A note on constants / well-known values and paths
+
+Somewhere we have to draw the line what should be configurable, what shouldn't and what should be hard-coded in the binary.
+
+We've decided to make the Kubernetes directory `/etc/kubernetes` a constant in the application, since it is clearly the given path in a lot of places,
```
dmmcquay: Suggested edit:

We've decided to make the Kubernetes directory `/etc/kubernetes` a constant in the application, since it is clearly the given path in a majority of cases,

docs/design/phases.md
```diff
@@ -0,0 +1,110 @@
...
+and the most intuitive location. Having that path configurable would for sure confuse readers of a on-top-of-kubeadm-implemented deployment solution.
+
+Further, having the admin/kubelet kubeconfig path would
```
dmmcquay: This doesn't make sense to me.

docs/design/phases.md
```diff
@@ -0,0 +1,110 @@
...
+     - the IP address of the default gateway
+     - optional extra altnames that can be specified by the user
+   - be in the `system:masters` organization
+ - a client certificate for the apiservers to connect to the kubelets securely (`apiserver-kubelet-client.crt`) using `ca.crt` as the CA with its private key (`apiserver-kubelet-client.key`). The certificate should:
+   - be a client certificate (`x509.ExtKeyUsageClientAuth`)
+   - be in the `system:masters` organization
+ - a certificate for signing ServiceAccount Tokens (`sa.crt`) using `ca.crt` as the CA with its private key (`sa.key`). The certificate should:
+   - ??
+
+If a given certificate and private key pair both exist, the generation will be skipped, and those files will validated and used if they are valid for the use-case.
```
dmmcquay: Suggested edit:

If a given certificate and private key pair both exist, the generation step will be skipped and those files will be validated and used for the prescribed use-case.

docs/design/phases.md
```diff
@@ -0,0 +1,110 @@
...
+This means the user can for example give his own CA which then will be used signing the rest of the certs.
```
dmmcquay: Suggested edit:

This means the user can, for example, give his own CA, which then will be used for signing the rest of the certs.

docs/design/phases.md
```diff
@@ -0,0 +1,110 @@
...
+ - a `/etc/kubernetes/tokens.csv` file that contains one kubeadm-generated token
+   - the content should be in the form `${token},${user},${uuid},${group}` where:
+     - `${token}` is the full token `123456.0123456789abcdef`
+     - `${user}` is currently `kubeadm-node-csr`, should be changed
+     - `${uuid}` is a random UUID created with `"k8s.io/apimachinery/pkg/util/uuid".NewUUID()`
+     - `${group}` is currently `kubeadm:kubelet-bootstrap`, might be changed
+
+### Bootstrap the control plane
+
+First, see so there is an etcd cluster somewhere:
```
dmmcquay: Suggested edit:

First, see that there is an etcd cluster somewhere:

docs/design/phases.md
```diff
@@ -0,0 +1,110 @@
...
+ - use an external one if external etcd params are given
+ - spin up one locally if external etcd should not be enabled. It should:
```
dmmcquay: Suggested edit:

spin up one locally if external etcd is not be enabled. It should:

docs/design/phases.md
```diff
@@ -0,0 +1,110 @@
...
+   - make a volume out to the host on a given host path
+
+Then, spin up the control plane as static pods:
+ - They are located in `/etc/kubernetes/manifests`
+ - Right now `.json` files, but will probably be `.yaml` files in v1.6
+ - `HostNetwork: true` is present on all Pods since there is no network configured yet
+
+Then kubeadm waits for the control plane to come up as static pods
+
+The master node is patched; the master label is added and the master is tainted.
```
dmmcquay: Suggested edit:

The master node is patched, the master label is added, and the master is tainted.

docs/design/phases.md
```diff
@@ -0,0 +1,110 @@
...
+### Configure the API Objects
+
+ServiceAccounts for `kube-dns` and `kube-proxy` are created in the `kube-system` namespace.
+
+If RBAC is enabled, then ClusterRoleBindings are created for:
+ - the TLS Bootstrap role (`system:node-bootstrapper`), so tokens in the group `kubeadm:kubelet-bootstrap` only can access what's required for TLS Bootstrapping
+ - kube-proxy; the `kube-proxy` ServiceAccount can gets the privileges in the `system:node-proxier` ClusterRole
+ - kube-dns; the `kube-dns` ServiceAccount can gets the privileges in the `system:kube-dns` ClusterRole
+
+In the future; the base layer of configuration for all kubelets in the cluster will be set in this phase.
```
dmmcquay: Suggested edit:

In the future the base layer of configuration for all kubelets in the cluster will be set in this phase.

luxas:
cc @mattymo @v1k0d3n @justinsb @chrislovecnm 
You're building tools on top of kubeadm; please comment here what you think about the planned, separately-invokable phases listed here

luxas:
cc @alexbrand for Kismatic as well :)

luxas:
cc @andrewrynhard

docs/design/phases.md
```diff
@@ -0,0 +1,110 @@
...
+ - a certificate for signing ServiceAccount Tokens (`sa.crt`) using `ca.crt` as the CA with its private key (`sa.key`). The certificate should:
+   - ??
+
+If a given certificate and private key pair both exist, the generation will be skipped, and those files will validated and used if they are valid for the use-case.
+This means the user can for example give his own CA which then will be used signing the rest of the certs.
+
+### Generate KubeConfig files on the master
+
+There should be:
+ - a KubeConfig file for kubeadm to use itself and the admin: `/etc/kubernetes/admin.conf`
```
philips: who is the admin?

docs/design/phases.md
```diff
@@ -0,0 +1,110 @@
...
+   - inside this file, a client certificate is generated from the `ca.crt` and `ca.key`. The client cert should:
+     - be a client certificate (`x509.ExtKeyUsageClientAuth`)
+     - be in the `system:nodes` organization
+
+The `ca.crt` is also embedded in the KubeConfig files.
+
+### Generate a token.csv file for bootstrapping purposes
+
+There should be:
+ - a `/etc/kubernetes/tokens.csv` file that contains one kubeadm-generated token
```
philips: gah, how did we end up with a CSV?

docs/design/phases.md
pires: It's a kubeconfig file for the cluster operator to copy to any machine where he wants to control the cluster from.

docs/design/phases.md
luxas: I know, I don't like this at all either, but this is what we've got for bootstrapping :/

docs/design/phases.md
mattymo: yeah, this is just an incomplete thought

docs/design/phases.md
```diff
@@ -0,0 +1,110 @@
...
+Certificates are stored by default in `/etc/kubernetes/pki`.
+
+There should be:
+ - a CA certificate (`ca.crt`) with its private key (`ca.key`)
+ - an API Server certificate (`apiserver.crt`) using `ca.crt` as the CA with its private key (`apiserver.key`). The certificate should:
+   - be a serving server certificate (`x509.ExtKeyUsageServerAuth`)
+   - contain altnames for
+     - the kubernetes' service's internal clusterIP and dns name (e.g. `10.96.0.1`, `kubernetes.default.svc.cluster.local`, `kubernetes.default.svc`, `kubernetes.default`, `kubernetes`)
+     - the hostname of the node
+     - the IP address of the default gateway
```
mattymo: why gateway IP? for snat tricks?

docs/design/phases.md
```diff
@@ -0,0 +1,110 @@
...
+     - optional extra altnames that can be specified by the user
```
mattymo: Can we include localhost and 127.0.0.1 in this list?

docs/design/phases.md
```diff
@@ -0,0 +1,110 @@
...
+   - be in the `system:masters` organization
+ - a client certificate for the apiservers to connect to the kubelets securely (`apiserver-kubelet-client.crt`) using `ca.crt` as the CA with its private key (`apiserver-kubelet-client.key`). The certificate should:
```
mattymo: one cert or one per client?

docs/design/phases.md
mattymo: substitute ; for ,

docs/design/phases.md
justinsb: (I'm not even sure you need to justify the choice of /etc/kubernetes)

docs/design/phases.md
```diff
@@ -0,0 +1,110 @@
...
+
+First off, kubeadm generates certificate and private key pairs for different purposes.
+Certificates are stored by default in `/etc/kubernetes/pki`.
+
+There should be:
+ - a CA certificate (`ca.crt`) with its private key (`ca.key`)
+ - an API Server certificate (`apiserver.crt`) using `ca.crt` as the CA with its private key (`apiserver.key`). The certificate should:
+   - be a serving server certificate (`x509.ExtKeyUsageServerAuth`)
+   - contain altnames for
+     - the kubernetes' service's internal clusterIP and dns name (e.g. `10.96.0.1`, `kubernetes.default.svc.cluster.local`, `kubernetes.default.svc`, `kubernetes.default`, `kubernetes`)
```
justinsb: Do you mean `100.96.0.1` ?  Or is it actually `10.96...`?

docs/design/phases.md
```diff
@@ -0,0 +1,110 @@
...
+     - the hostname of the node
```
justinsb: Presumably we aren't changing how kubeadm works here, so this is not the venue, but any notion of "hostname" is a huge red-flag for AWS compatibility

docs/design/phases.md
```diff
@@ -0,0 +1,110 @@
...
+     - the IP address of the default gateway
+     - optional extra altnames that can be specified by the user
+   - be in the `system:masters` organization
```
justinsb: This precludes/complicates use with an existing CA infrastructure I believe.  On the assumption though that doc is about the phases, and this is the existing kubeadm behaviour we are document, I'm going to stop calling out these issues.  If that's not right, and we are actually decided what the keys should look like, let me know and I will review with that mindset.

justinsb:
This looks cool.  kops could probably use the RBAC generation phase today (although we would want to be sure that it was HA safe, and that it did not reapply if users deleted the RBAC policies).  The RBAC layer feels almost like it could be a manifest - perhaps this meshes into the add-on-manager-v2 discussion (where I need to produce some form of strawman design doc based on the existing works).

luxas:
Thanks for the comments everyone, I'll look at this tomorrow or so and update it with your suggestion and more detail :+1:

docs/design/phases.md
luxas: Thanks for the grammatical nits, I was in a hurry when writing this so I didn't really pay attention to my wording in this doc, will fix :+1:

docs/design/phases.md
v1k0d3n: i'm with @philips on this one; why not json?

docs/design/phases.md
v1k0d3n: Suggested edit (nit-picky, only because it's been called out already):

going just a bit further: "First, determine if there is an etcd cluster available:"

docs/design/phases.md
luxas: @v1k0d3n the apiserver doesn't support reading tokens from json, so that's impossible for the time being

What we all want though, and what we're gonna implement is a dynamic way of managing these tokens so we can ditch this file totally (no one likes handling files on disk)

luxas: See https://kubernetes.io/docs/admin/kube-apiserver/
```
--token-auth-file string                                  If set, the file that will be used to secure the secure port of the API server via token authentication.
```

The new stuff is here: https://github.com/kubernetes/kubernetes/pull/41281

docs/design/phases.md
luxas: `10.96.0.1/12` is the default service subnet for kubeadm

docs/design/phases.md
```diff
@@ -0,0 +1,110 @@
...
+Somewhere we have to draw the line what should be configurable, what shouldn't and what should be hard-coded in the binary.
+
+We've decided to make the Kubernetes directory `/etc/kubernetes` a constant in the application, since it is clearly the given path in a lot of places,
+and the most intuitive location. Having that path configurable would for sure confuse readers of a on-top-of-kubeadm-implemented deployment solution.
+
+Further, having the admin/kubelet kubeconfig path would 
+
+### Certificates
+
+First off, kubeadm generates certificate and private key pairs for different purposes.
```
roberthbailey: Remove the "First off" part of the sentence.

docs/design/phases.md
roberthbailey: I think this should be optional and not included as a SAN by default.

docs/design/phases.md
roberthbailey: I don't think we want to have a cert signed for localhost. That seems like a security issue.

docs/design/phases.md
roberthbailey: That is an excellent point. We can't be too prescriptive about fields of the certs, since we need to allow certs generated by organizations to work smoothly with kubeadm.

docs/design/phases.md
roberthbailey: There's only one client (the apiserver).

docs/design/phases.md
```diff
@@ -0,0 +1,110 @@
...
+     - the kubernetes' service's internal clusterIP and dns name (e.g. `10.96.0.1`, `kubernetes.default.svc.cluster.local`, `kubernetes.default.svc`, `kubernetes.default`, `kubernetes`)
+     - the hostname of the node
+     - the IP address of the default gateway
+     - optional extra altnames that can be specified by the user
+   - be in the `system:masters` organization
+ - a client certificate for the apiservers to connect to the kubelets securely (`apiserver-kubelet-client.crt`) using `ca.crt` as the CA with its private key (`apiserver-kubelet-client.key`). The certificate should:
+   - be a client certificate (`x509.ExtKeyUsageClientAuth`)
+   - be in the `system:masters` organization
+ - a certificate for signing ServiceAccount Tokens (`sa.crt`) using `ca.crt` as the CA with its private key (`sa.key`). The certificate should:
+   - ??
```
roberthbailey: @mikedanese 

I think it needs to have the usage for signing certs (`KeyUsageCertSign`) and shouldn't have any other key usages set.

docs/design/phases.md
```diff
@@ -0,0 +1,110 @@
...
+
+If a given certificate and private key pair both exist, the generation will be skipped, and those files will validated and used if they are valid for the use-case.
+This means the user can for example give his own CA which then will be used signing the rest of the certs.
+
+### Generate KubeConfig files on the master
+
+There should be:
+ - a KubeConfig file for kubeadm to use itself and the admin: `/etc/kubernetes/admin.conf`
+   - inside this file, a client certificate is generated from the `ca.crt` and `ca.key`. The client cert should:
```
roberthbailey: I guess a client cert is ok here for bootstrapping, but once you have the cluster configured you should really throw it away in favor of configuring users so that things like audit logging can track who did what when.

docs/design/phases.md
```diff
@@ -0,0 +1,110 @@
...
+     - be a client certificate (`x509.ExtKeyUsageClientAuth`)
+     - be in the `system:masters` organization
+ - a KubeConfig file for kubelet to use: `/etc/kubernetes/kubelet.conf`
```
roberthbailey: If this kubelet is going to connect to the apiserver, shouldn't we use TLS bootstrapping to get it a client cert and give it a temporary token instead?

docs/design/phases.md
```diff
@@ -0,0 +1,110 @@
...
+     - `${uuid}` is a random UUID created with `"k8s.io/apimachinery/pkg/util/uuid".NewUUID()`
+     - `${group}` is currently `kubeadm:kubelet-bootstrap`, might be changed
+
+### Bootstrap the control plane
+
+First, see so there is an etcd cluster somewhere:
+ - use an external one if external etcd params are given
+ - spin up one locally if external etcd should not be enabled. It should:
+   - be spun up in a Pod
+   - listen on `localhost:2379`
```
roberthbailey: This will need to change once we support HA. In particular, we'll need to generate another set of certificates (possibly in a different cert universe) for the etcd instances to form a quorum.

docs/design/phases.md
```diff
@@ -0,0 +1,127 @@
...
+Nodes may also register themselves with the label `node-role.kubernetes.io/node=true` by using the `--node-labels` flag.
+
+The taint for proposed beta use is:
+
+```yaml
+apiVersion: v1
+kind: Node
+spec:
+  taints:
+  - key: node-role.kubernetes.io
```
luxas: @davidopp does this sound good to you?

luxas: cc @philips @jbeda

docs/design/phases.md
luxas: The usage of this file is gonna be replaced by the https://github.com/kubernetes/kubernetes/pull/36101 and https://github.com/kubernetes/kubernetes/pull/41281

docs/design/phases.md
pires: I know it's the default but do we really need a /12 (`10.96.0.0 - 10.111.255.255`) for this?

docs/design/phases.md
```diff
@@ -0,0 +1,127 @@
...
+Certificates are stored by default in `/etc/kubernetes/pki`.
+
+There should be:
+ - a CA certificate (`ca.crt`) with its private key (`ca.key`)
+ - an API Server certificate (`apiserver.crt`) using `ca.crt` as the CA with its private key (`apiserver.key`). The certificate should:
+   - be a serving server certificate (`x509.ExtKeyUsageServerAuth`)
+   - contain altnames for
+     - the kubernetes' service's internal clusterIP and dns name (e.g. `10.96.0.1`, `kubernetes.default.svc.cluster.local`, `kubernetes.default.svc`, `kubernetes.default`, `kubernetes`)
+     - the hostname of the node
+     - the IPv4 address of the default route
```
pires: What is this exactly, the apiserver pod IP?

docs/design/phases.md
```diff
@@ -0,0 +1,127 @@
...
+     - optional extra altnames that can be specified by the user
```
pires: I haven't seen this in the form of flags or configuration (haven't updated on the configuration API WIP PR). Are we planning ahead?

docs/design/phases.md
```diff
@@ -0,0 +1,127 @@
...
+   - ??be in the `system:masters` organization, this might not be needed??
```
pires: Is DNS integrated with RBAC in any way? I don't think so.

docs/design/phases.md
```diff
@@ -0,0 +1,127 @@
...
+ - a client certificate for the apiservers to connect to the kubelets securely (`apiserver-kubelet-client.crt`) using `ca.crt` as the CA with its private key (`apiserver-kubelet-client.key`). The certificate should:
+   - be a client certificate (`x509.ExtKeyUsageClientAuth`)
+   - be in the `system:masters` organization
```
pires: Is DNS integrated with RBAC in any way? I don't think so.

docs/design/phases.md
```diff
@@ -0,0 +1,127 @@
...
+ - a private key for signing ServiceAccount Tokens (`sa.key`) along with its public key (`sa.pub`)
+
+If a given certificate and private key pair both exist, the generation step will be skipped and those files will be validated and used for the prescribed use-case.
+This means the user can, for example, give his own CA, which then will be used for signing the rest of the certs.
+
```
pires: Can we add an extra paragraph stating it's the responsibility of the user to copy the CA (`ca.crt`) to whatever nodes need it for validating the generated certificates?

docs/design/phases.md
```diff
@@ -0,0 +1,127 @@
...
+ - a KubeConfig file for kubeadm to use itself and the admin: `/etc/kubernetes/admin.conf`
+   - the "admin" here is defined as `kubeadm` itself and the actual person(s) that is administering the cluster and want to control the cluster
+     - with this file, the admin has full control over the cluster 
+   - inside this file, a client certificate is generated from the `ca.crt` and `ca.key`. The client cert should:
+     - be a client certificate (`x509.ExtKeyUsageClientAuth`)
+     - be in the `system:masters` organization
+ - a KubeConfig file for kubelet to use: `/etc/kubernetes/kubelet.conf`
+   - inside this file, a client certificate is generated from the `ca.crt` and `ca.key`. The client cert should:
+     - be a client certificate (`x509.ExtKeyUsageClientAuth`)
+     - be in the `system:nodes` organization
```
pires: I still don't get why we need this but if it's just there for the sake of homogenization, I'm good.

docs/design/phases.md
pires: @luxas you should add your last comment to the document.

docs/design/phases.md
```diff
@@ -0,0 +1,127 @@
...
+     - `${uuid}` is a random UUID created with `"k8s.io/apimachinery/pkg/util/uuid".NewUUID()`
+     - `${group}` is currently `kubeadm:kubelet-bootstrap`, might be changed
+
+### Bootstrap the control plane
+
+First, determine if there is an etcd cluster available:
+ - use an external one if external etcd params are given
+ - spin up one locally if external etcd is not be enabled. It should:
+   - be spun up in a Pod
+   - listen on `localhost:2379`
```
pires: Please, add that it should be running in host network.

docs/design/phases.md
```diff
@@ -0,0 +1,127 @@
...
+   - set `PodSpec.SecurityContext.SELinuxOptions.Type=spc_t`
```
pires: Can you provide a link or a little explanation why?

docs/design/phases.md
```diff
@@ -0,0 +1,127 @@
...
+kind: Node
+spec:
+  taints:
+  - key: node-role.kubernetes.io
+    value: master
+    effect: NoSchedule
+```
+
+If self-hosting should be enabled, do a little bit more:
+ - TODO
```
pires: I should help here. When does this doc need to be merged?

docs/design/phases.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Configure the bootstrap tokens
+
+This phase configures the BootstrapSigner to create the public `cluster-info` ConfigMap in the `kube-public` namespace
```
pires: Can you please provide a link, even if to just the PR, to what the BootstrapSigner is?

docs/design/phases.md
```diff
@@ -0,0 +1,127 @@
...
+
+If RBAC is enabled, then ClusterRoleBindings are created for:
+ - the TLS Bootstrap role (`system:node-bootstrapper`), so tokens in the group `kubeadm:kubelet-bootstrap` only can access what's required for TLS Bootstrapping
+ - kube-proxy; the `kube-proxy` ServiceAccount can gets the privileges in the `system:node-proxier` ClusterRole
+ - kube-dns; the `kube-dns` ServiceAccount can gets the privileges in the `system:kube-dns` ClusterRole
+
+In the future, the base layer of configuration for all kubelets in the cluster will be set in this phase.
+
+Then the rest of the addons follow:
+ - apply the CNI Network Provider file if specified
```
pires: Please add, make sure CNI plug-in supports the new RBAC mode if RBAC is enabled.

docs/design/phases.md
```diff
@@ -0,0 +1,127 @@
...
+ - deploy kube-proxy as a DaemonSet
+   - the credentials to the master come from the ServiceAccount
+   - the location of the master comes from a ConfigMap
```
pires: We also need the pod network or is this gone now?

docs/design/phases.md
```diff
@@ -0,0 +1,127 @@
...
+
+Further, having the admin/kubelet kubeconfig path as constants as well (`/etc/kubernetes/admin.conf` and `/etc/kubernetes/kubelet.conf`) would make things easier as well.
+Is it ok to have `/etc/kubernetes/kubelet.conf` as a "standard" between official deployments? It defines the path for the kubelet's KubeConfig file
+
+### Certificates
+
+First off, kubeadm generates certificate and private key pairs for different purposes.
+Certificates are stored by default in `/etc/kubernetes/pki`.
+
+There should be:
```
deads2k: There should also be a CA keypair and client certificate for the Authenticating Proxy described here: https://kubernetes.io/docs/admin/authentication/#authenticating-proxy.  It's near zero cost in the kube-apiserver to have it, there's no exposure unless actual machine access is had, and future extension stories (https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/kubernetes-sig-api-machinery/o2v-HjAHk0k/Rg4lNeEACQAJ) rely upon it.

docs/design/phases.md
luxas: This has been possible from the beginning by using `--api-external-dns-names`, but it will probably be exposed in an other shape in the future (via the API Group)

docs/design/phases.md
luxas: This isn't related to DNS, but RBAC. Being in the `system:master` Organization (maps to Group in RBAC), basically makes you use `sudo` while talking to the cluster.

docs/design/phases.md
rothgar: Does kubectl read that file by default for system wide settings? I must have missed that and thought the config had to be in ~/.kube/

docs/design/phases.md
```diff
@@ -0,0 +1,127 @@
...
+Then, spin up the control plane as static pods:
+ - They are located in `/etc/kubernetes/manifests`
+ - The files are stored as `.yaml` files.
+ - `HostNetwork: true` is present on all Pods since there is no network configured yet
+
+Then kubeadm waits for the control plane to come up as static pods
+
+The master node is patched, the master label is added, and the master is tainted.
+
+The proposed master label for `v1.6` is `node-role.kubernetes.io/master=true`.
```
davidopp: Having a value of "true" is probably not useful. Why not just have the key (`node-role.kubernetes.io/master`) and no value?

docs/design/phases.md
davidopp: What is this taint supposed to signify? If it is for the master node(s) only, then use the same key and value (if any) as you use for the label on the master. (e.g. `node-role.kubenetes.io/master` with no value)

Also keep in mind that if you add a taint, then all pods that are supposed to run on the node will need a toleration for that taint.

docs/design/phases.md
pires: It makes so much sense to me that I now feel dumb because I didn't think about it first.

docs/design/phases.md
luxas: @pires Nodes should be in this group (organization), see: https://docs.google.com/document/d/1PqI--ql3LQsA69fEvRq1nQWgiIoE5Dyftja5Um9ML7Q/edit

docs/design/phases.md
luxas: This is what's returned from `0.0.0.0`, the IP address from the default (main) net interface

docs/design/phases.md
luxas: The ca.crt file is automatically transferred to the nodes thanks to our token bootstrap mechanism, and saved in /etc/kubernetes/pki/ca.crt by default

docs/design/phases.md
luxas: done

docs/design/phases.md
luxas: done

docs/design/phases.md
luxas: When the design is fully settled and agreed on :)

luxas:
Updated this now with a lot more detail about what should be done in kubeadm.
Also includes proposal for `kubeadm phase`

docs/design/phases.md
```diff
@@ -0,0 +1,354 @@
...
+ - `--tls-cert-file` to `apiserver.crt`
+ - `--tls-private-key-file` to `apiserver.key`
+ - `--kubelet-client-certificate` to `apiserver-kubelet-client.crt`
+ - `--kubelet-client-key` to `apiserver-kubelet-client.key`
+ - `--service-account-key-file` to `sa.pub`
+ - `--requestheader-client-ca-file` to `front-proxy-ca.crt`
+ - `--admission-control` to `NamespaceLifecycle,LimitRanger,ServiceAccount,PersistentVolumeLabel,DefaultStorageClass,ResourceQuota` or whatever the recommended set of admission controllers is at the time
+ - `--storage-backend` to `etcd3`. Support for `etcd2` in kubeadm is dropped.
+ - `--kubelet-preferred-address-types` to `InternalIP,ExternalIP,Hostname` so `kubectl logs` and other apiserver -> kubelet communication works in environments where the hostnames of the nodes aren't resolvable.
+ - `--requestheader-username-headers=X-Remote-User`, `--requestheader-group-headers=X-Remote-Group`, `--requestheader-extra-headers-prefix=X-Remote-Extra-`, --requestheader-allowed-names=front-proxy-client` so the front proxy communication is secure.
```
jbeda: Missing a "`" here.

docs/design/phases.md
```diff
@@ -0,0 +1,354 @@
...
+   - `--apiserver-port=6443`
+   - `--image-repository=gcr.io/google_containers`
+   - `--pod-network-cidr=`
+   - `--dns-domain=cluster.local`
+   - `--token=${AUTOGENERATE}`
+   - `--token-ttl=0`
+   - `--use-kubernetes-version=${STABLE_VERSION}`
+   - `--self-hosted`
+
+What would `kubeadm init` actually do?
```
jbeda: Not 100% necessary for v1 of this but it would be awesome if there were a kubeadm init --plan that would just output this for the options that are passed in. In this case users can understand what kubeadm is planning on doing in terms of the phases and make it clear that kubeadm is really just a bundle command.

docs/design/phases.md
```diff
@@ -0,0 +1,354 @@
...
+There should be:
+ - a CA certificate (`ca.crt`) with its private key (`ca.key`)
+ - an API Server certificate (`apiserver.crt`) using `ca.crt` as the CA with its private key (`apiserver.key`). The certificate should:
+   - be a serving server certificate (`x509.ExtKeyUsageServerAuth`)
+   - contain altnames for
+     - the kubernetes' service's internal clusterIP and dns name (e.g. `10.96.0.1`, `kubernetes.default.svc.cluster.local`, `kubernetes.default.svc`, `kubernetes.default`, `kubernetes`)
+     - the hostname of the node
+       - I guess this might be a requested feature in opinionated setups, but might be a no-no in more advanced setups. Consensus here?
+     - the IPv4 address of the default route
+     - optional extra altnames that can be specified by the user
```
timothysc: should describe how.  Ideally not just a massive command line arg list, but also a config file.

docs/design/phases.md
```diff
@@ -0,0 +1,354 @@
...
+### Bootstrap the control plane
+
+#### etcd
+
+Determine if the user specified external etcd options. If not, etcd should:
+ - be spun up as a Static Pod
+ - listen on `localhost:2379` and use `HostNetwork=true`
+ - have `PodSpec.SecurityContext.SELinuxOptions.Type=spc_t` set because of https://github.com/kubernetes/kubeadm/issues/107
+ - be at a minimum version `3.0.14`
+ - make a volume out to the host on a given host path
```
timothysc: We need to be certain we don't resource starve CPU on etcd + we need to make certain we open up ulimit caps ~ 65536 as a sane default.

docs/design/phases.md
```diff
@@ -0,0 +1,354 @@
...
+
+#### API Server
+
+The API Server needs to know this in particular:
+ - The subnet to use for services
+ - Where to find the etcd server
+ - The address and port to bind to; defaults to the IP Address of the default interface and port 6443 for secure communication
+ - Any extra flags and/or HostPath Volumes/VolumeMounts specified by the user
+
+Other flags that are set unconditionally:
```
timothysc: Is there a single mast.conf file we can create?  which may turn into a component.config

docs/design/phases.md
luxas: @timothysc Could you send a quick PR on the ulimit thing for etcd?

docs/design/phases.md
luxas: Everything described here will be available to config via the configfile which is going to expose a nice (the nicest) way of actually setting a kubeadm cluster up.
I should point that out much better...

docs/design/phases.md
luxas: ComponentConfig for the APIServer via a file isn't supported yet I think

cc @mtaufen

docs/design/phases.md
```diff
@@ -0,0 +1,354 @@
+## Phases in kubeadm
+
+`kubeadm init` does a lot. In order to make kubeadm the common building block for Kubernetes, each logical phase kubeadm does during `init` should be invokable separately. 
+Here's an explanation what kubeadm does in these different phases:
+
+#### A note on constants / well-known values and paths
+
+We have to draw the line somewhere about what should be configurable, what shouldn't, and what should be hard-coded in the binary.
+
+We've decided to make the Kubernetes directory /etc/kubernetes a constant in the application, since it is clearly the given path in a majority of cases,
```
bruceauyeung: I suggest /etc/kubernetes should be back-quoted

docs/design/phases.md
```diff
@@ -0,0 +1,354 @@
...
+
+
+If a given certificate and private key pair both exist, the generation step will be skipped and those files will be validated and used for the prescribed use-case.
+This means the user can, for example, give his own CA, which then will be used for signing the rest of the certs.
+
+### Generate KubeConfig files on the master
+
+There should be:
+ - a KubeConfig file for kubeadm to use itself and the admin: `/etc/kubernetes/admin.conf`
+   - the "admin" here is defined as `kubeadm` itself and the actual person(s) that is administering the cluster and want to control the cluster
```
bruceauyeung: I suggest : the "admin" here is defined as `kubeadm` itself and the actual person(s) who is administering the cluster and wants to control the cluster

docs/design/phases.md
```diff
@@ -0,0 +1,354 @@
+## Phases in kubeadm
+
+`kubeadm init` does a lot. In order to make kubeadm the common building block for Kubernetes, each logical phase kubeadm does during `init` should be invokable separately. 
+Here's an explanation what kubeadm does in these different phases:
+
+#### A note on constants / well-known values and paths
```
bruceauyeung: why h4 header here ? shouldn't it be h3 ?

docs/design/phases.md
```diff
@@ -0,0 +1,354 @@
...
+     - be a client certificate (`x509.ExtKeyUsageClientAuth`)
+     - have the name `system:kube-scheduler`
+
+`ca.crt` is also embedded in the KubeConfig files.
+
+### Bootstrap the control plane
+
+#### etcd
+
+Determine if the user specified external etcd options. If not, etcd should:
```
bruceauyeung: suggest: `kubeadm` determines if the user specified external etcd options in kubeadm config file.

bruceauyeung:
some suggestions

docs/design/phases.md
timothysc: So I dug on this and found out that we still don't support ulimit on pods, but we can override dockers startup defaults.. :-/

docs/design/phases.md
fabriziopandini: +1

fabriziopandini: Nice to have: Adding Aliases for each phase command with a numbered prefix could help users in understanding/using the right sequence of phases. e.g. 
> `kubeadm phase 1-certs ..` (where `1-certs` is alias of `certs`certificates`)
> `kubeadm phase 2-kubeconfig ..` (where `2-kubeconfig` is alias of `kubeconfig`)
Nb. this feature suites very well with autocompletion

docs/design/phases.md
```diff
@@ -0,0 +1,354 @@
...
+   - Outputs:
+     - Applies the built-in `kube-proxy` manifest to the API Server
+- `kubeadm phase addons kube-dns`
+   - Inputs:
+     - `--admin-kubeconfig=/etc/kubernetes/admin.conf`
+     - `--dns-domain=cluster.local`
+     - `--image-repository=gcr.io/google_containers`
+   - Outputs:
+     - Applies the built-in `kube-dns` manifest to the API Server
+
```
fabriziopandini: I suggest to provide a more consistent approach for outputs across the entire `kubeadm phase` UX, harmonising current approaches:
- `kubeadm phase certs selfsign` use `--cert-dir=`
- `kubeadm phase kubeconfig clientcert` writes to `stdout`
- `kubeadm phase controlplane *` writes to `/etc/kubernetes/manifests/*`
- `kubeadm phase addons cluster-info` apply directly to the apiserver
- eg

A first draft for possible alternative solution is:
- by default all `kubeadm phase` commands should provide the same experience of `kubeadm init`:
  - `kubeadm phase certs` and `kubeadm phase token` writes to `--cert-dir`, which defaults to `/etc/kubernetes/pki`
  - `kubeadm phase kubeconfig clientcert` writes to `/etc/kubernetes/; to be defined a rule for deriving filename from `--client-name=`(this is easier to implement if we will specialize kubeconfig phase)
  - `kubeadm phase controlplane *` writes to `/etc/kubernetes/manifests/*.yaml`
  - `kubeadm phase addons *` apply directly to the apiserver
- if invoked with a `--stdout`flag, all `kubeadm phase` commands should write to stdout; the output for the most of the commands is a yaml string or a sequence of kubectl commands; to be defined a format for `kubeadm phase certs`

docs/design/phases.md
```diff
@@ -0,0 +1,354 @@
...
+     - `--extra-args`
+   - Output:
+     - `kube-scheduler.yaml` in the manifests dir (`/etc/kubernetes/manifests`)
+ - `kubeadm phase addons cluster-info`
+   - Inputs:
+     - `--admin-kubeconfig=/etc/kubernetes/admin.conf`
+   - Outputs:
+     - Creates the `cluster-info` ConfigMap in the `kube-public` namespace
+     - Exposes it publicly with RBAC bindings
+- `kubeadm phase addons self-hosted`
```
fabriziopandini: might be that considering `self-hosted` as an `addons` can be confusing for users.
I suggest to treat `self-hosted` phase separately from `addons`

docs/design/phases.md
```diff
@@ -0,0 +1,354 @@
...
+     - `--alt-names=""`
+     - `--cert-dir=/etc/kubernetes/pki`
+   - Outputs (in the `--cert-dir` directory):
+     - `ca.{crt,key}`
+     - `apiserver.{crt,key}`
+     - `apiserver-kubelet-client.{crt,key}`
+     - `sa.{pub,key}`
+     - `front-proxy-ca.{crt,key}`
+     - `front-proxy-client.{crt,key}`
+ - `kubeadm phase kubeconfig clientcert`
```
fabriziopandini: Not sure if offering a generic `kubeconfig clientcert/token` phase can offer the simplest choice for users, because it requires to call this phase 4 times (kubelet,scheduler,controller-manager,admin), each one with very specific parameters.

A draft alternative proposal, might be specializing `kubeconfig` phase with dedicate sub-commands:
- `kubeconfig kubelet` (Inputs: `--api-server`, `--cert-dir`)
- `kubeconfig scheduler` (Inputs: `--api-server`, `--cert-dir`)
- `kubeconfig controller-manager` (Inputs: `--api-server`, `--cert-dir`)
- `kubeconfig user` (Inputs: `--api-server`, `--cert-dir`, `--name`, `--organization`)
All commands should produce a clientCert (default mode) expet if if `--token` flag is provided)

Additionally
- it could be useful to support a `kubeconfig all` (or simply `kubeconfig` without parameters) allowing to generate all 4 default config files in a single action
- especially for `kubeconfig user` it could be useful to to generate also an alternative output format as a `kubectl config set-cluster`, `kubectl config set credentials` commands and separate client files (e.g. useful for admin managing more that one cluster...)

docs/design/phases.md
```diff
@@ -0,0 +1,354 @@
...
+     - The kubeconfig is `stdout`
+ - `kubeadm phase kubeconfig token`
+   - Inputs:
+     - `--client-name=`
+     - `--api-servers=`
+     - `--token`
+     - `--cert-dir=/etc/kubernetes/pki`
+   - Output:
+     - The kubeconfig is `stdout`
+ - `kubeadm phase controlplane etcd`
```
fabriziopandini: it could be useful to support a `controlplane all` (or simply `controlplane` without subcommands) allowing to generate all control plane components in a single action

docs/design/phases.md
```diff
@@ -0,0 +1,354 @@
...
+   - Inputs:
+     - `--image=gcr.io/google_containers/etcd-${GOARCH}:3.0.14-kubeadm`
```
fabriziopandini: I suggest to keep an approach more consistent with `kubeadm init`, providing flags --image-repository=gcr.io/google_containers`and `--use-kubernetes-version=${STABLE_VERSION}` instead of `--image=...`

docs/design/phases.md
```diff
@@ -0,0 +1,354 @@
...
+     - `--data-dir=/var/lib/etcd`
```
fabriziopandini: By extending etcd flags, it is possible to provide with a minimal effort an early support for creating HA clusters, e.g.:
1. With `kubeadmin init`, create the first master with 
   - `--apiserver-advertise-address` and `--cert-altnames` properly set
   - an etcd.yaml that contains instructions for spinning up an etcd instance that should be member of an etcd cluster
2. Manually copy `/etcd/kubernetes` folder to a second and third masters (probably also necessary to replace kubelet.conf with a new one containing new credentials, created with `kubeconfig user`)
3. Manually setup a LB/DNS in front of all api-servers (matching with `--apiserver-advertise-address` and `--cert-altnames`)
4. Join nodes 

In order to support this a possibile solution is:
- add `--name` flag to etcd command, and assign to it a unique name (e.g. the nodename value retrived from Downward API)
- add a `--etcd-cluster-discovery` flag supporting a dynamic approach for setting up an etcd cluster (no need of knowing the name/ip of nodes in advance, but access to internet/to another etcd instance required)
- add a `--etcd-cluster-members' flag supporting a static approach for setting up an etcd cluster (required knowing the name/ip of nodes in advance, no need of access to internet/to another etcd instance)

docs/design/phases.md
```diff
@@ -0,0 +1,354 @@
...
+kubeadm phase controlplane controller-manager \
+	--image=${IMAGE_REPOSITORY}/kube-controller-manager-${ARCH}:${USE_KUBERNETES_VERSION} \
+	--cluster-cidr=${POD_NETWORK_CIDR} \
+	--cert-dir=${CERT_DIR}
+
+kubeadm phase controlplane scheduler \
+	--image=${IMAGE_REPOSITORY}/kube-scheduler-${ARCH}:${USE_KUBERNETES_VERSION}
+
+kubectl taint node $(hostname -i) node-role.kubernetes.io/master:NoSchedule
+kubectl label node $(hostname -i) node-role.kubernetes.io/master=""
```
fabriziopandini: Using `kubectl`in the middle of a `kubeadm`sequence of `phases`sounds a little bit strange. With low priority (but also low effort), it might make sense to implement a phase also for this step.

docs/design/phases.md
```diff
@@ -0,0 +1,354 @@
...
+kubectl taint node ${master_name} node-role.kubernetes.io/master:NoSchedule
+kubectl label node ${master_name} node-role.kubernetes.io/master=""
+```
+
+#### cluster-info
+
+This phase creates the `cluster-info` ConfigMap in the `kube-public` namespace as defined in https://github.com/kubernetes/community/blob/master/contributors/design-proposals/bootstrap-discovery.md
+ - The `ca.crt` and the address/port of the apiserver is added to the `cluster-info` ConfigMap in the `kubeconfig` key
+ - Exposes the `cluster-info` ConfigMap to unauthenticated users (i.e. users in RBAC group `system:unauthenticated`)
+
```
fabriziopandini: Looking at `kubeadm init` code, it seems there are also two additional steps here: `CreateServiceAccounts` and `CreateRBACRules`.
In order to make `kubeadm init` a full "bundle command", also those steps should be implemented as phases.

timothysc:
I'm of the opinion we should merge this and move along with updates as 1.8 progresses.

docs/design/design.md
```diff
@@ -0,0 +1,261 @@
+## Implementation design for kubeadm
+
+`kubeadm init` and `kubeadm join` together provides a nice user experience for creating a best-practice but bare Kubernetes cluster from scratch.
+However, it might not be obvious _how_ kubeadm does that.
+
+This document strives to explain what types of phases of work happen under the hood.
+Also included is ComponentConfiguration API types for talking to kubeadm programmatically.
```
mtaufen: Are you relying on the componentconfig API group? Be careful with this, that group will be deleted soon (as soon as every component's config has been moved out of it).

docs/design/design.md
```diff
@@ -0,0 +1,261 @@
...
+ - Easy to use
+   - The user should not have to run anything more than a couple of commands, including:
+     - `kubeadm init` on the master
+     - `export KUBECONFIG=/etc/kubernetes/admin.conf`
+     - `kubectl apply -f <network-of-choice.yaml>`
+     - `kubeadm join --token <token> <master>`
+     - The `kubeadm join` request to add a node should be automatically approved
+ - Extendable
+   - It should for example _not_ favor any network provider, instead configuring a network is out-of-scope
+   - Should provide a config file that can be used for customizing various parameters
```
mtaufen: Is this versioned?

mtaufen:
This is really cool!

docs/design/design.md
luxas: No, it's what I think the new ComponentConfiguration stuff looks like: https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/apis/kubeadm/v1alpha1/types.go

in it's own API group and everything. Not coupled to anything else, just reusing the API machinery

docs/design/design.md
luxas: yes, see the code above

luxas:
Thank you @mtaufen!

@timothysc Totally agree -- will merge now and follow up as we go.

docs/design/design.md
mtaufen: Ah, ok nice!

PR #180, opened 2017-02-23
Migrate bootkube checkpointer image.

dgoodwin:
This is just a raw copy of the checkpointer from https://github.com/kubernetes-incubator/bootkube/tree/master/cmd/checkpoint, with preserved git history.

I am proposing we first merge this in as is to make sure we get the history into the repo cleanly, then go to work on setting up the image build process.

CC @luxas @aaronlevy

### DIFF
--- a/cmd/checkpoint/README.md
+++ b/cmd/checkpoint/README.md
@@ -1,0 +1,74 @@
+# Checkpoint
+
+## Description
+
+`checkpoint` is a utility application which will manage "checkpoints" of pods scheduled to the local node.
+
+The purpose of a pod checkpoint is to ensure that existing local pod state can be recovered in the absence of an api-server.
+
+The kubelet will already attempt to ensure that local pods will continue to run in the absence of an API server (based on restartPolicy). However, if all local runtime state has been lost (for example after a reboot), the checkpoint utility will ensure the local state can be recovered until an api-server is contacted.
+
+This is accomplished by managing checkpoints as static pod manifests:
+
+- When the checkpointer sees that a "parent pod" (a pod which should be checkpointed), is succesfully running, the checkpointer will save a local copy of the manifest.
+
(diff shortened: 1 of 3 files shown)


luxas:
I'm ok with this as a starting point. LGTM

cc @mikedanese @jbeda -- feel the same?

dgoodwin:
On second thought maybe we should wait and bring it in one PR, I'm not sure we're going to like the build process looks.

Can anyone suggest what we want the build/make/vendoring to look like to get this to build? I'm not very familiar with go and it's dependencies so currently a bit lost in terms of how to make this build the binary, and then the image, but doing some digging. Please let me know if anyone has ideas.

aaronlevy:
I might second that - if we don't have a way to build the image, we're just going to end up maintaining code in two places. There are also some near term changes I wanted to get into the checkpointer (which we would have to double-commit if no images are being built here):

https://github.com/kubernetes-incubator/bootkube/issues/124
https://github.com/kubernetes-incubator/bootkube/issues/206
https://github.com/kubernetes-incubator/bootkube/issues/253

Just as another option - is there anything blocking from just using the images which are currently hosted on quay.io? If it's mostly an issue of annotations - thats something we can decide on a convention and just change.

aaronlevy:
Also something seems a bit odd with this PR - the actual history isn't associated with any of the code (click on one of the commits)

luxas:
@aaronlevy The direct blocker for us using the quay.io image is multi-platform support.
However, since we're not gonna make checkpointing support into v1.6 we have lots of time to think about this before we merge it into this repo.

@dgoodwin You could/should copy the scripts/Dockerfiles in `bootkube/images` or similar as well.
But we should probably switch over to `client-go` imports before building the binary for real.
When we have that, client-go will go into `cmd/checkpoint/vendor` or similar.

dgoodwin:
The process for preserving commit history involves a lot of black magic. I would suggest we forget about this, @aaronlevy if you or anyone in your group would like to just submit this file as is to get it into git, then we can go from there, that's ok with me. If not I propose we forget the history, and I can just redo this as a flat file copy.

dgoodwin:
Closing now, current plan to have bootkube build for required arches until we have build process in place for image in external repos like this one.

PR #236, opened 2017-04-11
Add proposal for local-cluster tool

ivan4th:
The proposal is mostly based on the current version of
[kubeadm-dind-cluster](https://github.com/Mirantis/kubeadm-dind-cluster).

The idea was previously discussed on sig-cluster-lifecycle (meetings, slack channel & during KubeCon).

The name of the tool can be changed if necessary if it's considered too confusing (`local-cluster` can replace `hack/local-up-cluster.sh` for many use cases but it's quite different)


ivan4th:
cc @luxas @errordeveloper @lukemarsden @zen @pigmej @idvoretskyi

docs/proposals/local-cluster.md
```diff
@@ -0,0 +1,486 @@
...
+CNI implementations are `weave`, `flannel` and `calico`.
+
+`NO_DEDICATED` (`--no-dedicated`, not set by default) removes
+`dedicated` taint from master node even if `NUM_NODES` is greater than
+zero.
+
+`DIND_SUBNET` (`--dind-subnet`, default: `10.192.0.0/16`) specifies a
+CIDR notation IP range for use by node containers.
+
+`APISERVER_PORT` (`--apiserver-port`, default: `8080`) specifies
```
luxas: hmm, we should use secure communication by default

docs/proposals/local-cluster.md
```diff
@@ -0,0 +1,486 @@
...
+apiserver port to use. If you're working with remote Docker engine,
+you need to forward this port to `localhost` over ssh.
+
+`BUILD_KUBEADM` (`--build-kubeadm`, not set by default) instructs
+`local-cluster` to build `kubeadm` locally from Kubernetes sources.
```
luxas: How do I specify which k8s dir?

docs/proposals/local-cluster.md
```diff
@@ -0,0 +1,486 @@
...
+sha1sum of `hyperkube` binary.
+
+`KUBECTL_URL` (`--kubectl-url`, not set by default) specifies
+the URL to download `kubectl` binary from. The binary will be used
+on user's machine and must match its OS and architecture.
+
+`KUBECTL_SHA1` (`--kubectl-sha1`, not set by default) specifies
+sha1sum of `kubectl` binary.
+
+`LOCAL_GO` (`--local-go`, not set by default) instructs
```
luxas: This is not a supported way of compiling kubernetes. We should _always_ do the containerized build

docs/proposals/local-cluster.md
```diff
@@ -0,0 +1,486 @@
...
+`local-cluster` to use local Go installation to build
+`kubeadm`/`hyperkube`/`kubectl`. By default `local-cluster` uses k8s
+build container for this task. _NOTE:_ implementating this option may
+make the startup script more complex.
+
+`NO_DASHBOARD` (`--disable-dashboard`, not set by default) instructs
```
luxas: Instead of `NO_{ADDON}` and `{ADDON}_URL` we should have `ADDONS=file1.yaml,file2.yaml` etc.

docs/proposals/local-cluster.md
```diff
@@ -0,0 +1,486 @@
...
+be used to set up a cluster on a GCE instance that's prepared using
+`docker-machine`.
+
+## Implementation notes
+
+### Base image
+
+`local-cluster` by default uses a prebuilt image to start the cluster.
+There's an image per each supported Kubernetes version and supported
+architecture (`amd64`, `arm`, `arm64`). There's also an image for each
```
luxas: now also ppc64le and s390x

docs/proposals/local-cluster.md
```diff
@@ -0,0 +1,486 @@
...
+
+The following local filesystem layout is used by `local-cluster`
+(particular `kubectl` binary versions are given as an example):
+
+```
+  $HOME
+    |
+    +-- .local-cluster/
+              |
+              +--- config.sh
```
luxas: should be `config.yaml` with an API specific to `local-cluster-up`

docs/proposals/local-cluster.md
```diff
@@ -0,0 +1,486 @@
...
+pulls the binaries directly from build container. _NOTE:_ extra tricks
+may be necessary in case `LOCAL_GO` is enabled.
+
+### Networking
+
+`local-cluster` utilizes named docker network
+`k8s-local-cluster`. This network is always created or replaced upon
+the cluster startup using address range specified via `DIND_SUBNET`
+configuration variable. `local-cluster` makes docker's internal DNS
+server available to all the nodes in the cluster, so nodes can be
```
luxas: node name being routable with DNS isn't necessary and most often not the case in users' setup

docs/proposals/local-cluster.md
```diff
@@ -0,0 +1,486 @@
...
+The snapshot is made by means of preserving each node's
+`/var/lib/docker` directory as a docker volume and saving each node
+container's filesystem in an image. The volumes and images created by
+snapshotting mechanism are removed during `./local-cluster.sh clean`.
+
+### Future directions
+
+The following features may be added later:
+
+* automatic PV creation like it's done in `minikube`, so most Helm
```
luxas: this is not easily doable since this is a multinode cluster.

However, we can use Rook or the like easily to get such functionality

docs/proposals/local-cluster.md
```diff
@@ -0,0 +1,486 @@
...
+this, there should be a test that builds Kubernetes binaries based on
+`master` branch of
+[kubernetes/kubernetes](https://github.com/kubernetes/kubernetes)
+repository. k8s `master` test should be executed daily so possible
+issues that break DIND can be tracked down in timely manner and with
+minimal use of `git bisect`.
+
+### Limitations
+
+There are problems with `btrfs` support because of a
```
luxas: Can we just state that `overlay` and `overlay2` are the supported drivers? (and maybe `aufs`, does that work?)

philips:
Idea for another direction to take to achieve the goal of control plane load testing: use a single machine/kubelet setup (e.g. laptop, minikube, one vm, etc) but launch a few more [kubemark](http://blog.kubernetes.io/2016/07/kubernetes-updates-to-performance-and-scalability-in-1.3.html) type pods to simulate additional load.

mikedanese:
@pipejakob @krousey @roberthbailey

roberthbailey:
@ivan4th - can you address comments from @luxas and then put this onto the agenda for the next cluster lifecycle meeting?

ivan4th:
Yes, sure, sorry for the delay, I worked on  [Virtlet update](https://github.com/Mirantis/virtlet/pull/261) for k8 1.6 during last week, which involved some tedious debugging sessions. I'll update the document tomorrow well before the meeting and add it to the agenda. The main changes I'm making are these:
- don't use starter script, there'll be `local-cluster-up` binaries for Linux and Mac
- consequently, no more control container, `local-cluster-up` will do everything
- don't manage kubectl binaries for the user
- use secure communication with apiserver
- remove support for `LOCAL_GO`
- use proper `.yaml` file for config (although env vars will be supported, too)
- ppc64le and s390x archs
- more clean addon management
- add an example of config file

kargakis:
Why is this specific to the kubeadm repo? Having the ability to spin up multiple nodes is something that core devs can take advantage of (I certainly have felt the need of such a feature with hack/local-up-cluster). Why not switch hack/local-up-cluster under the hood to use kubeadm+dind?

docs/proposals/local-cluster.md
ivan4th: From my experience having fixed node names like `kube-master`, `kube-node-1`, ... etc is much more convenient for use then looking up/remembering node IPs. E.g. for things like `kubectl label node kube-node-1 ...`

docs/proposals/local-cluster.md
ivan4th: It's also possible to make flexvolume driver that uses docker volumes.

docs/proposals/local-cluster.md
ivan4th: Well, `kubeadm-dind-cluster` seems also to work ok when host docker has `devicemapper` storage.

docs/proposals/local-cluster.md
ivan4th: Will fix this

docs/proposals/local-cluster.md
ivan4th: Ok, although it'll probably require some effort on my side to set up VMs for these.

docs/proposals/local-cluster.md
ivan4th: A bit unclear because e.g. the dashboard is usually installed from an url and not local yaml.

docs/proposals/local-cluster.md
ivan4th: Ok will remove this.

docs/proposals/local-cluster.md
ivan4th: It defaults to current directory (will document this) but I'll also add an option to specify it.

docs/proposals/local-cluster.md
ivan4th: Ok will fix

docs/proposals/local-cluster.md
```diff
@@ -0,0 +1,486 @@
...
+```
+$ wget https://cdn.rawgit.com/kubernetes/kubeadm/master/local-cluster/local-cluster-v1.6.sh
+$ chmod +x local-cluster-v1.6.sh
+
+$ # start the cluster
+$ ./local-cluster-v1.6.sh
+
+$ # make it easier to access the cluster
+$ export KUBECONFIG=$HOME/.local-cluster/admin.conf
+$ alias kubectl=$HOME/.local-cluster/kubectl
```
ivan4th: Ok, will remove kubectl management

docs/proposals/local-cluster.md
```diff
@@ -0,0 +1,486 @@
...
+binaries for different supported platforms? Also, with control
+container the cluster can still start successfully if the connection
+between user's machine and docker engine breaks. Concerning multiple
+image versions with different binaries included: the idea is to be
+able to start cluster on user's machine without downloading much data
+once the image is pulled.
+
+## Usage
+
+`local-cluster.sh` startup script defaults to start the most recent
```
ivan4th: Ok, as my starter script idea seems to be not very good after all, let's provide `local-cluster-up` binaries for Linux and Mac, will update the doc.

docs/proposals/local-cluster.md
```diff
@@ -0,0 +1,486 @@
...
+These use cases are actually already covered by
+[kubeadm-dind-cluster](https://github.com/Mirantis/kubeadm-dind-cluster). The
+plan is to adapt this project for inclusion in `kubeadm` repository.
+
+## Proposal
+
+The `local-cluster` tool consists of a docker image that contains
+Docker engine and main `local-cluster` binary and a small startup
+script that starts the control container and downloads proper version
+of `kubectl` to the user's machine. There are several image versions
```
ivan4th: Ok, removing kubectl management

ivan4th:
@kargakis `hack/local-up-cluster.sh` is not DIND and also is quite easily customizable in every detail which may be important in some contexts, e.g. for testing some CRI impls, although e.g. in [Virtlet](https://github.com/Mirantis/virtlet) we managed to live quite well with kubeadm and `kubeadm-dind-cluster` thanks to Virtlet's CRI proxy mechanism. BTW I think there's possibility to use host docker instead of inner one, along the lines of [kxd](https://github.com/errordeveloper/kxd) project.

BTW @philips mentioned there can be problems with running kubelet in DIND env, could you please elaborate a bit? Perhaps it has to do with resource management? Because for the purpose of working on k8s and projects that extend `kubeadm-dind-cluster` is quite ok from my experience.

errordeveloper:
@kargakis good point indeed. The reason why we suggested to @ivan4th to work within this repos is primarily to do with the fact that this tool would an enabler for kubeadm development. Replacing `hack/local-up-cluster.sh` sounds like a great idea to me, but I don't think code would need to live in main repo for that to happen, we can consider separate repo, but to me this one would be just fine.

errordeveloper:
@philips I don't think load-testing was covered here, and I would leave that out for the moment, as we really need to focus on making it easier and quicker for anyone to run integration tests, by the way, this is really a lot about contributor experience (ping @kubernetes/sig-contributor-experience-proposals).

errordeveloper:
cc @kubernetes/sig-contributor-experience-proposals

docs/proposals/local-cluster.md
```diff
@@ -0,0 +1,422 @@
...
+  charts from
+  [kubernetes/charts](https://github.com/kubernetes/charts) repository
+  work out of the box
+* enabling feature gates for Kubernetes components
+* automatic installation of `helm` and `tiller`
+* support using `delve` to debug k8s components.  Should be an option
+  for building k8s with `GOGCFLAGS="-N -l"` plus `delve` binary in the
+  images. Possible problem: as of Go 1.7, trying to build e.g.
+  `kubectl` with `GOGCFLAGS="-N -l"` may result in stack overflow
+  in Go compiler.
```
errordeveloper: @ivan4th could you please add what @philips suggested about simulating extra load with kubemark to this list...

luxas:
@ivan4th as of the sig meeting, do you mind moving this to kubernetes/community?

Thanks!

ivan4th:
@nhlfr Yes, I also thought about such an abstraction, but I think it can be implemented at later stages so as to simplify the initial implementation.

ivan4th:
Closing in favor of kubernetes/community#629

PR #286, opened 2017-06-02
Add triage doc

### LINKED ISSUE #251: Create process for triaging/fixing kubeadm test failures
Now that we have continuously running [end-to-end tests](https://k8s-testgrid.appspot.com/google-gce#gce-kubeadm), and the desire to [add more](https://github.com/kubernetes/kubeadm/issues/190), we should create a shared process that sets expectations for community members around who should triage and fix kubeadm, or the tests themselves, when they start failing.

Here are some of the scenarios that have already come up and could have benefited from a clear policy:

- A contributor creates a legitimate kubeadm bug and e2es immediately start failing.
  - Seems like they should fix their bug after tests start failing (time/availability permitting).
- A contributor changes kubeadm behavior in an expected way that works, but breaks our current testing.
  - Should they be held responsible for always updating our e2e tests for their changes? We do expect people to update unit tests.
- An upstream change to the `test-infra` code (e.g. `bootstrap.py`, `kubernetes_e2e.py`, etc.) breaks the kubeadm end-to-end tests.
  - Hopefully they would provide a fix if they were notified about the regression, but we don't have a clear contract between SIGs around this, and will need buy-in from the EngProd team for any expectations we have for their support.

This is one of the action items from the [1.6.0 postmortem](https://docs.google.com/document/d/1o-wJ39O8eqBxOCHgu0MGmgH4EdojzI3WlrsV3TVwhe0/edit?usp=sharing).

jamiehannaford:
Fixes #251. 

@luxas suggested this might also be submitted to kubernetes/community. Before doing so, do folks think that 

1. this doc can be abstracted,
2. it's relevant enough to other SIGs?

### DIFF
--- a/docs/test-failures.md
+++ b/docs/test-failures.md
@@ -1,0 +1,44 @@
+# Triaging and fixing kubeadm test failures
+
+Kubeadm now has a series of end-to-end tests which run for every pull request
+and merge action against the repo. Tests, however, often fail.
+
+This document seeks to create a shared process that sets expectations for
+community members around who should triage and fix kubeadm, or the tests
+themselves, when they start failing.
+
+## A new feature is submitted which inadvertently breaks correct tests
+
+If tests begin to fail on a pull request which introduces new functionality or
+refactors old, and the tests are still deemed relevant, it is the responsibility
+of the contributor to fix the broken tests.
+
+This process will tie into the usual feedback cycle for pull requests: if a PR
+build fails, the CI process will inform the contributor of which tests are failing,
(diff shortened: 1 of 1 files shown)


luxas:
ping @roberthbailey @pipejakob @lukemarsden @jbeda PTAL

roberthbailey:
/lgtm 

Let's definitely merge it into our repo, and then we can offer it to the community repo and see if they'd like to adopt it as well (with slightly modified language, since some parts of the project don't have a brittle upstream while others do).

luxas:
Merging, we can follow-up as needed...

PR #391, opened 2017-08-15
Update Code of Conduct

mbssaiakhil:
Made the last sentence uniform wrt other sentences


pipejakob:
I assume this file was just copied from the main `kubernetes/kubernetes` repository originally. Since the upstream file has been updated to simply point to the CNCF Code of Conduct, maybe we should just copy over the new version of the file instead? I don't think that `kubeadm` has a distinct code of conduct compared to the rest of Kubernetes.

https://github.com/kubernetes/kubernetes/blob/master/code-of-conduct.md

luxas:
That's right and very reasonable @pipejakob 
@mbssaiakhil If you want to proceed with this PR, please do what Jakob suggested.

mbssaiakhil:
Thanks @pipejakob and @luxas but I have a doubt.

The code of conduct file in https://github.com/kubernetes/kubernetes/blob/master/code-of-conduct.md was last changed on Oct 19 2016, whereas in kubeadm it was added on Nov 22 2016 in commit (https://github.com/kubernetes/kubeadm/commit/1ce93d7f0910712abcaf87e84f41f1df4a54b140). 

The file at https://github.com/kubernetes/kubernetes/blob/master/code-of-conduct.md which is linked to (i.e https://github.com/cncf/foundation/blob/master/code-of-conduct.md) is also not same as the code of conduct which is in kubeadm i.e https://github.com/cncf/foundation/blob/master/code-of-conduct.md. 

So, it looks like, we are maintaining a different file for code of conduct in kubeadm. Kindly suggest.

luxas:
cc @dankohn @sarahnovotny @bgrant0607 
I don't think that is/was expected.

(we badly need automation for this...)

dankohn:
Yes, @luxas https://github.com/kubernetes/kubeadm/blob/master/code-of-conduct.md should either be:

1. Removed. It is clear from the org that kubeadm is part of Kubernetes and so inherits the Kubernetes code of conduct.
2. Changed to be a reference to https://github.com/kubernetes/kubernetes/blob/master/code-of-conduct.md

I would suggest 1 but either option is fine.

luxas:
Thanks @dankohn!
I think option 1 is the best one, as it doesn't add extra overhead regarding keeping things in sync over `n` different repos. Very good if this automatically applies to the other repos :+1:

@mbssaiakhil Can you please just remove this file then as discussed?
Thanks!

mbssaiakhil:
Nice, it would be one less file to maintain. Can the similar concept be applied to other repos as well? Can I go ahead and remove this file from other repos as well, so that every repo would have one less file to maintain?

Btw I have removed the code-of-conduct.md in the PR. @luxas kindly review and merge

pipejakob:
Actually, can you just squash this into a single commit? Then I'll merge. Thanks!

mbssaiakhil:
Thanks @pipejakob, I have squashed the commits. Kindly review and merge.

PR #499, opened 2017-10-17
Add CoC

jamiehannaford:
Just realised we don't have a Code of Conduct doc for this repo. Adding so our community expectations are explicit ✨

### DIFF
--- a/code-of-conduct.md
+++ b/code-of-conduct.md
@@ -1,0 +1,3 @@
+## Kubernetes Community Code of Conduct
+
+Kubernetes follows the [CNCF Code of Conduct](https://github.com/cncf/foundation/blob/master/code-of-conduct.md).


luxas:
Thank you @jamiehannaford for the initiative, but please see https://github.com/kubernetes/kubeadm/pull/391 (I had no idea about this policy before either)
Not sure how well the rest of the repos follow this pattern...

jamiehannaford:
Based on what @dankohn said in [this comment](https://github.com/weaveworks/weave/issues/3016#issuecomment-321337923), it seems like we do have the option to have a CoC if it points to the canonical CNCF version. IMO it's worth having a CoC per repo because:

- new members who begin contributing to this repo (and/or those who continue to do so) might not know where to go to see the "real" CoC. Having it here removes the ambiguity.
- being as explicit as possible about our community expectations is important because it means nobody can feign ignorance by saying "Oh I didn't see anything in this repo, I just assumed you didn't have one".
- other projects were open to adding them (see https://github.com/kubernetes/minikube/pull/1579, https://github.com/kubernetes/dashboard/pull/2026, https://github.com/kubernetes/heapster/pull/1261).

I also don't think having a 2-3 line file is an unreasonable maintenance burden in this scenario. Just my 2c 😄  I would like to hear what other folks think.

luxas:
oops, commented right away without looking at the content :wink:. sorry about that.
Having a link SGTM, for the reasons mentioned above, but not maintaining our own copy.

/lgtm

Can we get someone else to sign off as well?

jamiehannaford:
@luxas Due to lack of feedback, can we merge this?

PR #575, opened 2017-11-28
add rpms script

jamiehannaford:
/cc @luxas


jamiehannaford:
@luxas Looks like `v1.6.10 v1.6.9 v1.6.8` are missing rpms and debs. Should we exclude them from tests or add them somehow?

PR #704, opened 2018-02-16
Update troubleshooting for create cluster failure

jessicaochen:
Provide guidance on where to look for common failure cases for kubeadm

PR #723, opened 2018-03-06
Cleanup formatting in test-failures.md

jessicaochen:
Followup cleanup from PR https://github.com/kubernetes/kubeadm/pull/704

/cc @roberthbailey @xiangpengzhao

### DIFF
--- a/docs/test-failures.md
+++ b/docs/test-failures.md
@@ -48,8 +48,9 @@ will always be brittle - so regular communication is a must.
 There is an issue in provisioning the cluster.
 
 Test failure will look like:
+
 ```error during make -C /workspace/kubernetes-anywhere WAIT_FOR_KUBECONFIG=y deploy: exit status 2```
 
 To debug further, dig into the collected logs
-1. Determine if master was able to set up with kubeadm. Logs are under: /artifacts/master-node-name/serial-1.log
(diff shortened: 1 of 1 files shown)


roberthbailey:
Thanks!

docs/test-failures.md
```diff
@@ -48,8 +48,9 @@ will always be brittle - so regular communication is a must.
 There is an issue in provisioning the cluster.
 
 Test failure will look like:
+
 ```error during make -C /workspace/kubernetes-anywhere WAIT_FOR_KUBECONFIG=y deploy: exit status 2```
 
 To debug further, dig into the collected logs
-1. Determine if master was able to set up with kubeadm. Logs are under: /artifacts/master-node-name/serial-1.log
-2. Determine if nodes were able to set up with kubeadm. Logs are under: /artifacts/node-name/serial-1.log
+1. Determine if master was able to set up with kubeadm. Logs are under: ```/artifacts/master-node-name/serial-1.log```
```
philoserf: for a singler line only a single tick is needed

```
this is `code` and the rest is not
```
this is `code` and the rest is not

-----

1. Determine if master was able to set up with kubeadm. Logs are under: `/artifacts/master-node-name/serial-1.log`

xiangpengzhao: @philoserf this makes sense. But since triple ticks works in this case as well, I think we can leave it as-is. WDYT?

PR #795, opened 2018-05-09
kubeadm-contrib-ux

fabriziopandini:
As discussed in Cophenanghen, the contrib experience for kubeadm should be improved, supporting a consistent approach to manual testing during each release cycle.

This PR extends the first version of the kubeadm vagrant test environment from @stealthybox by adding the capability to create a set of vagrant machines starting from a cluster API spec.

IMPORTANT! This PR focuses only on the final user experience and it is meant to get feedbacks on the proposed solution; all the necessary hack/ansible scripts will follow in a subsequent PRs

/CC @timothysc @luxas @stealthybox @liztio @detiber @chuckha


luxas:
@fabriziopandini ping me and Tim when this is ready for a final review/approve.
@stealthybox please work with Fabrizio to take this out of the PoC stage

vagrant/README.md
```diff
@@ -1,39 +1,48 @@
...
-git checkout feature/kubeadm_594-etcd_tls
-bazel test //cmd/kubeadm/...
-bazel build //cmd/kubeadm --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64
-issue=594 ~/go/src/k8s.io/kubeadm/vagrant/copy_kubeadm_bin.sh
-
-git checkout feature/kubeadm_710-etcd-ca
-bazel test //cmd/kubeadm/...
-bazel build //cmd/kubeadm --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64
-issue=710 ~/go/src/k8s.io/kubeadm/vagrant/copy_kubeadm_bin.sh
-```
```
stealthybox: This documentation describing `bazel test` and `bazel build` is very valuable.
It shows new contributors the development workflow.
We shouldn't remove it or build additional tools around these commands.

vagrant/README.md
```diff
@@ -1,39 +1,48 @@
...
-#### experiment with the two builds on the vagrant:
-```shell
-cd ~/go/src/k8s.io/kubeadm/vagrant
-
-vagrant up
-vagrant ssh
-  sudo /vagrant/bin/594_kubeadm init
-  sudo /vagrant/bin/594_kubeadm reset
-  sudo /vagrant/bin/710_kubeadm init
-```
```
stealthybox: We should provide an example of how to compare builds with the new workflow since operating on the same cluster with two builds of kubeadm is the most common edge case we deal with.

vagrant/Vagrantfile
```diff
@@ -1,32 +1,62 @@
...
+        setupControlplaneVip: true,               # this action is performed only if there are 2 or more machines with Master role  (see ClusterAPI, MachineSets/Role Master)
+        kubeadmInit: false, 
+        applyNetwork: false,
+        kubeadmJoinMaster: false,                 # this action is performed only if there are 2 or more machines with Master role  (see ClusterAPI, MachineSets/Role Master)
+        kubeadmJoin: false                        # this action is performed only if machines with Node role are defined            (see ClusterAPI, MachineSets/Role Node)
+    )
+    #.injectKubeadmFromBazelBuild(prefix: nil)    # uncomment if you want to inject a kubeadm release from bazel build; set the prefix if you want to differentiate the binary name
+    #.injectKubeadmFromDockerBuild(prefix: nil)   # uncomment if you want to inject a kubeadm release from docker build; set the prefix if you want to differentiate the binary name
+    #.injectKubeadmFromLocalBuild(prefix: nil)    # uncomment if you want to inject a kubeadm release from local build; set the prefix if you want to differentiate the binary name
```
stealthybox: It needs to be possible to load these builds into `/vagrant` out of band from the `vagrant up`.
A dev shouldn't have to reprovision the vm when they make a bugfix or a one-line logging change.
In the previous workflow, the vm is never really recreated because overwriting the build artifact and running `kubeadm reset` is generally enough to get you back to a point where you can rerun tests.

Can we use `vagrant provision --provision-with <..>` for this?
One thing that's nice about the shell script is that it can easily be run from the kubernetes source dir (outside the vagrant directory).

vagrant/Vagrantfile
```diff
@@ -1,32 +1,62 @@
...
+        v.name = "#{machineSpec.name}"
+  	    v.memory = "#{machineSpec.memory}"
+        v.cpus = "#{machineSpec.cpus}"
+        v.customize ["modifyvm", :id, "--uartmode1", "disconnected"]
+        v.linked_clone = true if Vagrant::VERSION =~ /^1.8/
+      end
+
+      if machineSpec.name == clusterSpec.machines.last.name
+        machine.vm.provision :ansible do |ansible|
+          ansible.compatibility_mode = "2.0"
```
stealthybox: This looks like a neat option -- does it allow us to care less about what ansible version we depend on?

vagrant/spec/README.md
```diff
@@ -0,0 +1,93 @@
...
+          defines the target DNS addon of choice>
+        kubeletConfig:
+          <systemdDropIn|dynamicKubeletConfig (optional, default systemdDropIn)
+          defines the target DNS addon of choice>
+        networkAddon:
+          <weavenet|flannel|calico (optional, default weavenet)
+          defines the target network add-on of choice>
+      etcd:
+        version: <string (optional, default v3.2.17), defines the version of etcd binaries to be installed - only extenal etcd - >
+```
```
stealthybox: I really like how you're using Cluster API for this.
It allows us to maintain a similar experience if we recreate this workflow for other tools such as kubeadm-dind-cluster.

It might also be possible to use other Cluster API controllers with this dev-tooling as well.

vagrant/spec/masters.yaml
```diff
@@ -0,0 +1,18 @@
...
+  template: 
+    metadata: 
+    spec: 
+      providerConfig: 
+        value: 
+          apiVersion: vagrantMachineSet/v1alpha1
+          kind: vagrantMachineSet
+          box: "ubuntu/xenial64"
+          cpus: 2
+          memory: 2048
```
stealthybox: These seem like good CPU/mem defaults

One thing that's notable is that this is a base-image change.
The previous workflow uses `bento/ubuntu-17.10` which is a Chef community maintained box for Artful Aardvark along with the kubernetes xenial repos.

timothysc:
@fabriziopandini when you have a moment could you address @stealthybox 's comments.

vagrant/README.md
fabriziopandini: +1, done

vagrant/README.md
fabriziopandini: done

vagrant/Vagrantfile
fabriziopandini: done.
Now `kubeadm-playground create` and `kubeadm-playground deploy` are separated commands. There is also a paragraph describing how to handle iterative dev cycles

vagrant/Vagrantfile
fabriziopandini: Yes, it avoids some ansible warning and force you tu use latest formats for inventory file and so on

vagrant/spec/README.md
fabriziopandini: totally agree. 
Now the ansible part and the vagrant part are even more decoupled, and it will be easier to introduce e.g. controllers or other tools like terraform as alternative to vagrant

vagrant/spec/masters.yaml
fabriziopandini: That's a simple example/default configuration.
You can change this one or better create your own cluster api specification to pass to `kubeadm playground` command

fabriziopandini:
@stealthybox all the comment are addressed
@timothysc @luxas If you take a look at the `readme.md` file you can can quickly catch up the idea behind this proposal; basically it is a `kubeadm-playground` command with following options:

```bash
start     Starts a new kubeadm playground
deploy    Deploy kubeadm binary from build output to playground machines
ssh       Ssh into one playground machine
exec      Executes ansible playbooks for changing the state of playground (e.g. reset)
e2e       Executes ansible playbooks implementing E2E tests sequences
stop      Stops the current kubeadm playground
delete    Delete the current kubeadm playground
status    Return the status of the machines in the current playground
help
```

timothysc:
This is -devel side so merging.  Folks can PR follow ups.

PR #796, opened 2018-05-09
kubeadm-contrib-hack-lib

fabriziopandini:
This PR implements the second part of https://github.com/kubernetes/kubeadm/pull/795 (2 of 3), that consists of a set of scripts that are responsible for parsing the cluster API spec and translates it into something that can be used by Vagrant and by the ansible provisioner.

NB. the test guide is only drafted, and it is intended to be improved over time ;-)

/CC @timothysc @luxas @stealthybox @liztio @detiber @chuckha


luxas:
cc @kris-nova ^

Issue #991, opened 2018-07-13
kubeadm-playground exits failing OS detection on Mac /w ansible 2.3.3.0 


stealthybox:
## Versions
**kubeadm repo branch:**
  0840c14900073b74a59891fb53c85a5587314124 (master)

**ansible:**
```
❯ ansible --version
ansible 2.3.3.0
  config file = /Users/leigh.schrandt/.ansible.cfg
  configured module search path = Default w/o overrides
  python version = 2.7.14 (default, Mar 22 2018, 15:04:47) [GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)]
```

**OS:**
MacOS Sierra
Darwin Kernel Version 16.7.0: Thu Jun 21 20:07:39 PDT 2018; root:xnu-3789.73.14~1/RELEASE_X86_64 x86_64

## What happened?
The devtool fails with this message when trying to provision a new dev cluster with the default spec:
```
~/go/src/k8s.io/kubeadm/vagrant/hack/ansible master
❯ ../../kubeadm-playground start

KUBEADM PLAYGROUND [Checking machine status] **********************************
All the machines are already running

KUBEADM PLAYGROUND [Preparing for ansible execution] **************************

KUBEADM PLAYGROUND [Run ansible] **********************************************

PLAY [install ansible prerequisites] **********************************************************************************************************************************************

PLAY RECAP ************************************************************************************************************************************************************************
kubeadm-test-master        : ok=1    changed=0    unreachable=0    failed=0

ERROR! no action detected in task. This often indicates a misspelled module name, or incorrect module path.

The error appears to have been in '/Users/leigh.schrandt/go/src/k8s.io/kubeadm/vagrant/hack/ansible/roles/common/tasks/main.yml': line 2, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

---
- import_tasks: debian.yml
  ^ here


The error appears to have been in '/Users/leigh.schrandt/go/src/k8s.io/kubeadm/vagrant/hack/ansible/roles/common/tasks/main.yml': line 2, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

---
- import_tasks: debian.yml
  ^ here


PLAY [creates the kubeadm config] *************************************************************************************************************************************************

TASK [kubeadm-config : gets the kubeadm version] **********************************************************************************************************************************
[...]

PR #994, opened 2018-07-16
Change ansble use 'include' instead 'import_tasks' or 'include_tasks'

chenyb4:
Signed-off-by: Yuanbin.Chen <cybing4@gmail.com>

### DIFF
--- a/vagrant/hack/ansible/roles/common/tasks/main.yml
+++ b/vagrant/hack/ansible/roles/common/tasks/main.yml
@@ -1,6 +1,6 @@
 ---
-- import_tasks: debian.yml
+- include: debian.yml
   when: ansible_os_family == "Debian"
 
-- import_tasks: redhat.yml
+- include: redhat.yml
   when: ansible_os_family == "RedHat"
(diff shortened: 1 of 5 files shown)


stealthybox:
Related to issue #991 

The `include` module is deprecated in newer versions of ansible but currently still functions:
> This module will still be supported for some time but we are looking at deprecating it in the near future.
https://docs.ansible.com/ansible/2.6/modules/include_module.html

This does resolve the problem though.

kad:
With ansible 2.6 out, and with possibility to use ansible from virtualenvs, I'd suggest not to use deprecated module for just support ansible 2.3 or lower. I'd rather set expectations about supported ansible versions.

fabriziopandini:
@chenyb4 sorry for delay in reviewing this
I agree with @kad the we should't use deprecated modules, so my proposal is to close this PR

FYI  
I'm going to raise at SIG level a discussion about the future of kubeadm-playground which is quickly moving from a rough idea discussed at kubeconf to something bigger. In the long term we should simplify things to make it more easily maintenable and getting rid of ansible dependency is something to tackle asap

PR #996, opened 2018-07-16
Add instructions on updating command line reference docs

### LINKED ISSUE #893: explain how we submit PRs for generated command docs in CONTRIBUTING.md
the submission for `k/website` PRs about things like kubeadm commands has to be explained in detail, so that we can get other folks working on documenting new commands.

right now i believe only me and @fabriziopandini know how to do that from the kubeadm team.

the addition should be made in the `CONTRIBUTING.md` file in the `k/kubeadm` repo.

@timothysc @Bradamant3

neolit123:
Fixes #893


docs/updating-command-reference.md
```diff
@@ -0,0 +1,39 @@
...
+documentation is generated automatically. The generated output can be found here:
+* https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm/
+
+If a command or a command flag is added or removed this has to be reflected in the documentation on a new release.
+The `kubernetes/website` repository has the tooling to facilitate this, but some manual steps are still required.
+
+#### Example scenario: adding a new sub-command
+
+- Implemented a new kubeadm configuration sub-command called `kubeadm config newcommand`.
+- Run `./hack/update-generated-docs.sh`. This will generate new files in the folder `./docs/admin/` called `*newcommand*`.
```
fabriziopandini: If I remember well we should use `hack/generate-docs.sh` instead

docs/updating-command-reference.md
```diff
@@ -0,0 +1,39 @@
...
+- If you are sending a PR to `kubernetes/kubernetes` make sure that you add these files
```
fabriziopandini: Before adding files to k/k you should run `hack/update-generated-docs.sh`. it replace cobra helps with placeholders

neolit123:
@fabriziopandini thanks for the review.

[`./hack/update-generated-docs.sh`](https://github.com/kubernetes/kubernetes/blob/master/hack/update-generated-docs.sh) is basically [`./hack/generate-docs.sh`](https://github.com/kubernetes/kubernetes/blob/master/hack/generate-docs.sh) but it also replaces with placeholders.

right now, we commit placeholders in both `k/k` and `k/w` as both repos have tooling to generate docs on release and the full generated docs are not needed.

so, i think that my explanation in the doc is valid, please correct me if i'm wrong again.
(applies to both review comments).

fabriziopandini:
@neolit123 for sure you have more updated info than me (I was not aware k/w generates doc automatically)

However it might be useful for the readers to add a note explaining this e.g.
"Please note that ./hack/update-generated-docs.sh generate doc placeholders, because the real doc will be generated by k/w during the build process; if you want to see a preview of the generated doc use  ./hack/generate-docs.sh instead"

WDYT?

neolit123:
> "Please note that ./hack/update-generated-docs.sh generate doc placeholders:"

@fabriziopandini i agree, makes sense to add a note about that.

neolit123:
@fabriziopandini 
hi, just added the extra note to the document:
https://github.com/kubernetes/kubeadm/pull/996/files#diff-1937753e1778c9f2a37bfdab8540c0fbR41

fabriziopandini:
Great work!
/lgtm

PR #1010, opened 2018-07-19
Disable swap in all nodes

aojea:
`kubeadm` fails with swap on https://github.com/kubernetes/kubernetes/pull/31996

Added tasks to disable swap in all nodes from https://github.com/kubernetes/contrib/blob/master/ansible/roles/node/tasks/swapoff.yml


riverzhang:
/lgtm  Kubeadm Playground , I want to join it.
@timothysc  @fabriziopandini

fabriziopandini:
/lgtm
/approve

@riverzhang happy to have you on board! Feel free to contact me on slack!

PR #1139, opened 2018-09-22
Update README.md

AdamDang:
Line 98: If you are developing new kubeadm futures...
I guess here should be kubeadm features, right?

### DIFF
--- a/vagrant/README.md
+++ b/vagrant/README.md
@@ -95,7 +95,7 @@ And then execute your kubeadm test scenario (See following paragraphs for more i
 
 ## Iterative dev/build/release/test cycles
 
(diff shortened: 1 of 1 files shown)


vagrant/README.md
```diff
@@ -287,4 +287,4 @@ Then you can connect to the playground with `kubeadm-playground ssh` and:
 
 In order to make working `kubeadm-playground` provisioning faster, you can create a
 vagrant base box with all the prerequisites already in place.
-see [instruction](box.md) for more info
\ No newline at end of file
+see [instruction](box.md) for more info
```
dixudx: Does this line get changed? They are exactly the same. Did I miss something?

PR #1144, opened 2018-09-25
move verify_packages_published to it's own folder

neolit123:
also:
- update the test README to reflect the folder change.
- add a README in e2e/tests explaining that this folder
contains e2e tests and have some links in there.
- do some minor whitespace / comment cleanup in the .sh

my idea here is to have folders under e2e that specify e2e type.
(having this as early as possible is better!)

tests/e2e/packages is what would be available now.
tests/e2e/manifests is what's coming soon.

test-infra PR on hold here:
https://github.com/kubernetes/test-infra/pull/9565

/assign @fabriziopandini 

### DIFF
--- a/tests/e2e/README.md
+++ b/tests/e2e/README.md
@@ -1,12 +1,8 @@
-## Tools for verifying Kubernetes packages are published
+### kubeadm e2e tests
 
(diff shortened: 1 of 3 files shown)


neolit123:
/check-cla
hm..

PR #1148, opened 2018-09-27
e2e-docs: instruct about removing old x-on-y jobs

neolit123:
as per the discussion in slack with @timothysc, old x-on-y jobs should be removed too, while old upgrade jobs should be left.

this PR amends the e2e doc.

/kind cleanup
/assign @timothysc 

edit: PR for test-infra will follow.

### DIFF
--- a/docs/managing-e2e-tests.md
+++ b/docs/managing-e2e-tests.md
@@ -1,4 +1,3 @@
-
 ## Managing kubeadm end-to-end tests
 
 ### Overview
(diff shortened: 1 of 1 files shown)


docs/managing-e2e-tests.md
```diff
@@ -1,4 +1,3 @@
-
```
neolit123: ^ small whitespace cleanup.

PR #1157, opened 2018-10-04
add an e2e test for verifying manifest lists

neolit123:
this PR adds an e2e test for verifying the manifests lists.
on top of `verify_manifest_lists.go` there are some details on how this works.

QUESTION: this basically traverses manifest lists json files until it reaches the image layers.
currently download of the layers is skipped and only the `Content-Length` of their headers is verified.
these can end up being quite large, so should the full download of all layers be enabled? (the toggle is called `downloadLayers`).

test-infra PR to follow.

/hold
/assign @timothysc 
/cc @dims @BenTheElder @ixdy @Random-Liu

### DIFF
--- a/tests/e2e/manifests/verify_manifest_lists.go
+++ b/tests/e2e/manifests/verify_manifest_lists.go
@@ -1,0 +1,697 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+verify_manifest_lists.go
+
+This program contains tests for Docker schema 2, multi-arch manifest lists
+that kubeadm requires for every release.
+
+First, it tries to get the latest Kubernetes release tags from GitHub.
+The list of release tags is filter so that it doesn't contain versions
+that are either too old or are known to not support manifest lists.
+
+It then creates a map of required images and image tags by looking
+at the constants defined in the kubeadm source code for a release
+branch estimated from a release tag.
+
+Once the images and image tags are defined, the program starts downloading
+the manifest lists from GCR, but also the actual images and layers,
+while verifying their contents and sizes. The layer download can be
+skipped by toggling `downloadLayers` to false.
+
+Results are cached so that a certain "image:tag" doesn't have to be verified
+more than once.
+*/
+
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"sort"
+	"strconv"
+	"strings"
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/version"
+)
+
+const (
+	// optional download of image layers.
+	downloadLayers = false
+
+	// default timeout for HTTP requests in seconds.
+	defaultHTTPTimeout = 10
+
+	// first release that is known to have full support.
+	firstKnownVersion = "v1.12.0-rc.1"
+
+	gcrBucket = "https://gcr.io/v2/google-containers"
+
+	typeManifestList = "application/vnd.docker.distribution.manifest.list.v2+json"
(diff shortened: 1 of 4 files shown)


ixdy:
@timothysc this is in the kubeadm repo, not kubernetes, so it has nothing to do with the existing kubernetes e2e tests as best as I can tell.

timothysc:
@ixdy hahaha - thanks.  I didn't even notice, I review too many PRs in a day.

ixdy:
(to be fair I missed that the first time too)

neolit123:
> @neolit123 Umm this is not really a standard e2e. It's a wrapped tool, that we'll need to manage..?.?

some goes for https://github.com/kubernetes/kubeadm/tree/master/tests/e2e/packages
it's just a test that tests existence of artifacts.

@timothysc if you have maintenance concerns and that this should not be placed here, i'm going to ask if another repo wants to shelter it.

if there are no k8s takers, i will place it in a github repo under my account, but then it would not run as a periodic job.

tests/e2e/manifests/verify_manifest_lists.go
```diff
@@ -0,0 +1,674 @@
...
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"sort"
+	"strconv"
+	"strings"
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/version"
```
ixdy: is this vendored in this repo? I don't see it...

ixdy:
does this maybe belong in core, though? manifest lists + multiarch isn't really kubeadm-specific, is it?

tests/e2e/manifests/verify_manifest_lists.go
```diff
@@ -0,0 +1,674 @@
...
+	images["kube-proxy"] = k8sVersionV
+	images["etcd"] = ""
+	images["pause"] = ""
+	images["coredns"] = ""
+	// TODO(neolit123): kube-dns is being deprecated eventually [*]
+	images["k8s-dns-kube-dns"] = ""
+	images["k8s-dns-sidecar"] = ""
+	images["k8s-dns-dnsmasq-nanny"] = ""
+
+	// parse the constants file and fetch versions
```
ixdy: this seems a bit fragile. is there a better way to query or list the necessary packages and versions?

tests/e2e/manifests/verify_manifest_lists.go
neolit123: the `verify_manifest_lists.sh` script only pulls the `version` package instead of the whole apimachinery repo.
i didn't want to do `go get` or `dep`, because it only needs a couple of files in reality.
if there are objections to this i can change it.

tests/e2e/manifests/verify_manifest_lists.go
neolit123: sadly, no.
given the test targets kubeadm primary, there is not other location where the tags for images are defined, other than the `constants.go` file in kubeadm.

neolit123:
@ixdy 
> does this maybe belong in core, though? manifest lists + multiarch isn't really kubeadm-specific, is it?

extracting `constants.go` from the kubeadm tree makes is kubeadm centric though. i don't think we have a k8s location where all the tags for all the images for a certain release are defined?

if we do, then we can possibly extend this to check the manifest lists for all the images bound to a certain k8s release.

neolit123:
/lint

- fixed some typos.
- updated to go 1.11.1.
- ran locally: typecheck, misspell, go vet, go fmt, golint.

timothysc:
/assign @ixdy @dims  

Could you two lgtm or agree on a happy home?  It's a separate tool, but it's use is broader then kubeadm. I'm happy to land it here for now and we could move it to a new home when needed.

neolit123:
right now, the tool can be extended to verify an "extra list" of images that also follow the k8s versioning, but kubeadm doesn't need:
```
cloud-controller-manager:v1.12.0
hyperkube:v1.12.0
...?
```

is there a way to see the full list of such images?

but it won't be possible for the tool to know how to verify images that don't follow the k8s versioning - e.g. `fluentd-elasticsearch`.

this can be tricky, because the tool or potential consumers would need a JSON payload to be defined somewhere that has k8s -> image-x mappings.

neolit123:
/lgtm cancel

the 1.13.0-alpha.1 was just cut without the existence of a release-1.13 branch.
this breaks the logic of this test.

it thinks that it would make sense for the branch to exists when a release is cut...but nope...
will fix it soon.

timothysc:
@ixdy ^

timothysc:
I'm unblocking this due to timeout.

neolit123:
ok, thanks.
in that case i can create a job that can be viewed in test-grid.

PR #1187, opened 2018-10-26
Update box.md

AdamDang:
There are 3 presentations for "pre-pull" in this doc:
Line 10:pre-pull,
Line 28: pre pull, 
Line 33: prepull
Better to use the same format: pre-pull.


fabriziopandini:
/approve
/lgtm
Thanks

PR #1313, opened 2018-12-08
WIP: enable terraform as deploy-tool on libvirt-kvm

MalloZup:
This pr is a proposal to use in addition to vagrant the terraform-libvirt provider which i'm co-maintainting  on https://github.com/dmacvicar/terraform-provider-libvirt.

I proposing it as wip for gathering some ideas and interest first before i do the project.

I wil follow the same structure as `vagrant` and tooling etc.

We could do it similar to https://github.com/MalloZup/ceph-open-terrarium/tree/master/examples/libvirt with all distro supported in longterm.

### DIFF
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -1,0 +1,3 @@
+# Kubeadm Playground (terraform)
+
+The kubeadm playground consists of one or more terraform libvirt  machines where you can experiment kubeadm and eventually test a local build of kubeadm itself.


neolit123:
hi @MalloZup , you need to sign the CLA if you want to contribute to k8s.
i have to admit that i'm personally not a big fan of adding documentation for a terraform playground, but i will defer to others on options about this. :)

/kind feature
@kubernetes/sig-cluster-lifecycle-pr-reviews

MalloZup:
@neolit123 thank you for feedback! :sunflower:  So i signed the CLA, should be ok.

I'm fine with both decisions. I think is just to up to the team to choose. I agree it is not super-important but just a secondary feature, based also on workflows.

Thx :rocket:

MalloZup:
@fabriziopandini @fabriziopandini  ok i got the point. Thank you for answers. We can close it.

My goal was to use `terraform` instead of  of `vagrant` as contributor, but if we plan to use `kind` i'm fine with it. Thanks for links! cu :sunflower:

PR #1389, opened 2019-02-06
kind: add simple test-cmd script and single-cp config

neolit123:
Add a "basic" test-cmd script to be used from the
kind deployer.

Also include a single-cp config for kind.

Mostly experimenting at this point.
The directory structure is a subject to change
once more test cases are added.

test-infra job PR is here: TODO

### DIFF
--- a/tests/e2e/kind/test-cmd-basic.sh
+++ b/tests/e2e/kind/test-cmd-basic.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+# Copyright 2019 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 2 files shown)


tests/e2e/kind/test-cmd-basic.sh
```diff
@@ -0,0 +1,50 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# export the KUBECONFIG
+KUBECONFIG="$(kind get kubeconfig-path)"
+export KUBECONFIG
+
+# base kubetest args
+KUBETEST_ARGS="--provider=skeleton --test --check-version-skew=false"
```
timothysc: why not just use sonobuoy?

timothysc:
I'm not really a fan of the scriptery, but it looks like you're doing this for automation.

neolit123:
> I'm not really a fan of the scriptery, but it looks like you're doing this for automation.

hopefully this .sh can be reused by a multitude of tests.
the alternative is to write it in `go`.

PR #1396, opened 2019-02-08
Local environment setup improvement

bowiesaurus:
2 changes here to improve getting set up for kubeadm-playground development.

1. Added a stock `Pipfile` with the required `pyyaml` package and Python version to make local development less fraught with missing dependencies.

2. Updated the alias for `kubeadm-playground` to make sure it allows `pipenv` to pick up on the correct path to the `Pipfile`.

### DIFF
--- a/vagrant/Pipfile
+++ b/vagrant/Pipfile
@@ -1,0 +1,12 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
+pyyaml = "*"
+
+[dev-packages]
+
+[requires]
+python_version = "2.7"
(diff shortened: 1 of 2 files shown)


yagonobre:
Hi @bowiesaurus, thanks for you pr. Can you add a description to this? Also if you want to contribute to kubenetes you need to sign the CLA

neolit123:
the kubeadm playgound is deprecated and we probably shouldn't accept more PRs against it.
/hold

bowiesaurus:
@neolit123 This is what we were able to find just to get started on setting up kubeadm development for what we needed at the time. If there is a better, well-documented workflow elsewhere, linking it here would be much appreciated.

neolit123:
the playground is being deprecated in favor of this tool:
https://github.com/kubernetes-sigs/kind

the docs for it are here:
https://github.com/kubernetes-sigs/kind/tree/master/docs/user

all you need is the kubernetes source tree and Docker 18.00 and you can create a test environment.

please have a look and let me know if it fits your user case.

neolit123:
please re-open if needed.

PR #1446, opened 2019-03-12
Welcome kinder!

fabriziopandini:
kinder is an example of [kind](https://github.com/kubernetes-sigs/kind) used as a library, designed for simplifying kubeadm development/local testing

This PR introduces the kinder roadmap/readme file, for collecting the first round of feedback about goals/high-level features before adding the actual kinder code

/kind documentation
/priority important-longterm

/assign @timothysc 
/assign @BenTheElder 
/assign @neolit123 

/cc @kubernetes/sig-cluster-lifecycle-pr-reviews

PR #1450, opened 2019-03-13
kinder

fabriziopandini:
after https://github.com/kubernetes/kubeadm/pull/1446 merged, this PR adds the actual kinder code

/kind feature
/priority important-longterm

/assign @timothysc 
/assign @neolit123 


go.mod
```diff
@@ -0,0 +1,3 @@
+module k8s.io/kubeadm
+
+go 1.12
```
neolit123: hm, i though that it is possible to have the module under the kinder folder
e.g. `module k8s.io/kubeadm/kinder` ?

neolit123: oh, there is one already...
in which case do we need this one?

fabriziopandini: it is automatically generated, so I think it is needed...

neolit123: isn't it possible to run `go module` only under the `kinder` folder and not the `/` folder?
or is this required for go 1.12 to work like so?

fabriziopandini: if I use go modules in the kinder folder, go automatically generated go.mod in this folder and in the parent folder. It is required/managed by go

neolit123: ok

.vscode/settings.json
```diff
@@ -0,0 +1,17 @@
...
+        "Wrapf",
+        "kcfg",
+        "kcluster",
+        "kctx",
+        "logrus",
+        "nginx",
+        "nslookup",
+        "srcs"
+    ]
+}
```
ereslibre: I would tend to not add specific IDE settings if possible.

kinder/OWNERS
```diff
@@ -0,0 +1,9 @@
+# See the OWNERS file documentation:
+#  https://github.com/kubernetes/community/blob/master/contributors/devel/owners.md
+approvers:
+- fabriziopandini
+- neolit123
+reviewers:
```
yagonobre: Do you need to add you and @neolit123 as reviewer as well?

fabriziopandini: Good point. I will do this immediately after this one merge

neolit123:
@timothysc mentioned that he is on PTO this week.
as per our previous agreement to have this tool as a replacement for kubeadm-playground i'm going to cancel the hold.

if we face complications and objections post merge we can discuss and amend.

/hold cancel

PR #1482, opened 2019-04-02
add kinder get artifacts command

fabriziopandini:
This PR adds the new `kinder get artifacts` that allows to retrive Kubernetes build artifacts from different sources

This is a useful feature and also an initial step for implementing kinder build node-variant from build artifacts, that is a piece of the puzzle for getting E2E tests on kubeadm upgrades

/kind feature
/priority important-longterm

/assign @neolit123

### DIFF
--- a/kinder/cmd/kinder/get/get.go
+++ b/kinder/cmd/kinder/get/get.go
@@ -1,0 +1,48 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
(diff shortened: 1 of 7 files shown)


kinder/cmd/kinder/get/get.go
```diff
@@ -0,0 +1,48 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package get implements the `get` command
+package get
```
neolit123: we add get.go for artifacts, but does it makes sense to have the other subcommands too?
(the build of this file might fail because king/get/* packages are missing).

fabriziopandini: having the kinder list of commands implemented as a superset of the kind list of commands IMO is nice because it avoids users to switch back and forth from the two commands.

the risk is that something will break when we will update the kind dependency, but this applies to all the "kind as a library" surface as well because it is something not yet stable and well defined, so I'm fine with this until kinder is experimental and we can change if required

neolit123: actually, i got confused that we are using go modules and didn't know where the "get" sub-packages are coming from if we are not vendoring the kind cmd package.

but your comment makes sense.
given upstream is aiming for a stable CLI soon, i think that such extensions will be safe, for now.

/lgtm

PR #1506, opened 2019-04-14
kinder/hack: add verify and update scripts

neolit123:
/assign @fabriziopandini 
/priority important-longterm
/kind feature

please review one commit at a time.
i will setup the presubmit CI for this repo next week.


kinder/hack/update-deps.sh
```diff
@@ -0,0 +1,56 @@
...
+    -not -iname "CONTRIBUTORS*" \
+    -not -iname "COPYING*" \
+    -not -iname "LICENSE*" \
+    -not -iname "NOTICE*" \
+    -exec rm '{}' \;
+}
+
+export GO111MODULE="on"
+go mod tidy
+# TODO enable in kinder?
```
fabriziopandini: Unless necessary, I think we should wait until kinder gets more stable (end of this cycle) before starting to vendor, so we can change without getting PR big.

kinder/hack/update-deps.sh
```diff
@@ -0,0 +1,56 @@
...
+# go mod vendor
+# prune-vendor
+
+# TODO(bentheelder): re-enable once (if?) we're using docker client
```
fabriziopandini: I don't think this apply to kinder

kinder/hack/update-deps.sh
neolit123: ack.

kinder/hack/update-deps.sh
neolit123: updated.

neolit123:
/hold
looks like there is a bug in update-deps.sh.

one concern here is that with the k8s publishing bot the deps update daily for api-machinery.

PR #1530, opened 2019-04-23
kinder: update calico manifest to latest stable

mauilion:
We should consider how we track this going forward.
With every release of calico the url will change.

A potential solution is to track github.io/kube-flannel.yaml instead as
there are fewer moving parts and it will track master in the flannel
repo.

Signed-off-by: Duffie Cooley <dcooley@heptio.com>


mauilion:
@neolit123 @fabriziopandini

PR #1536, opened 2019-04-25
kinder: upgrade stable-to-master refinements

fabriziopandini:
This PR adds a couple of refinements to upgrade stable-to-master
- documents where the job configuration and dashboard are
- changes variables in order to
  - use labels for stable / master versions
  - sets baseImage, image, clusterName with default + allows override via env var
- build the test image from tars 
- sets timeouts to avoid flakes
- adds logging to kinder get logs (that is the flaking)

In order to make this work, some minor adjustments were made to the waiter (to handle properly build releases) and to the template engine (to allow the usage of env var only as an override) 

Important: this should merge only after https://github.com/kubernetes/kubernetes/pull/77072

/assign @neolit123


kinder/hack/verify-all.sh
```diff
@@ -13,6 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# IMPORTANT! This script is called by pull-kubeadm-kinder-verify presubmits job
+# see https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/kubeadm/kubeadm-presubmits.yaml
```
neolit123: :+1:

fabriziopandini:
@neolit123 
Removed env var/defaut as suggested

neolit123:
thanks for the update.
/lgtm

PR #1539, opened 2019-04-25
kinder: add support for include workflow

fabriziopandini:
This PR adds support for including workflow

This enables the reuse of the same list of task in several workflows (as an alternative to make workflows configurable via env variables)

/assign @neolit123

### DIFF
--- a/kinder/pkg/test/workflow/workflow.go
+++ b/kinder/pkg/test/workflow/workflow.go
@@ -20,9 +20,12 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"path/filepath"
+	"regexp"
 	"time"
 
 	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
 
 	yaml "gopkg.in/yaml.v2"
 )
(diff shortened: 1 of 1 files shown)


kinder/pkg/test/workflow/workflow.go
```diff
@@ -137,11 +148,77 @@ func NewWorkflow(file string) (*Workflow, error) {
...
+				continue
+			}
+			log.Warnf("env var %s in workflow file %s is shadowed by env var %[1]s in parent workflow file %[3]s", k, path, file)
+		}
+
+		// import all tasks from the import file into the parent file, removing task name prefix
+		re := regexp.MustCompile(`^task\-\d{2}\-?`)
+		for _, tx := range wx.Tasks {
+			tx.Name = re.ReplaceAllString(tx.Name, "")
+			w.Tasks = append(w.Tasks, tx)
```
neolit123: is it adding the tasks of the imported workflow at the bottom of the already existing tasks of the parent workflow?

fabriziopandini: It is replacing the import task with tasks in the imported workflow (not adding at the bottom)
 e.g.

workflow A
- Task 1 
- Task 2 import workflow B
- Task 3 

And

workflow B
- Task 1 
- Task 2 

Will became
- Task 1 (from A/task1)
- Task 2 (from B/task1)
- Task 3 (from B/task2)
- Task 4 (from A/task2)

fabriziopandini: You can think at this as an include

neolit123: thanks for the example
/lgtm

PR #1542, opened 2019-04-27
kinder: task ignoreError

fabriziopandini:
This PR adds support for setting tasks to ignore errors

This allows to makes test pass if kind get logs flakes but everything else pass

/assign @neolit123

### DIFF
--- a/kinder/ci/workflows/upgrade-stable-master.yaml
+++ b/kinder/ci/workflows/upgrade-stable-master.yaml
@@ -134,6 +134,10 @@ tasks:
     - "{{ .env.ARTIFACTS }}"
   force: true
   timeout: 5m
+  # kind export log is know to be flaky, so we are temporary ignoring errors in order
+  # to make the test pass in case everything else passed
+  # see https://github.com/kubernetes-sigs/kind/issues/456
+  ignoreError: true
 - name: reset
   description: |
     Exec kubeadm reset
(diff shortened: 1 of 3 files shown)


kinder/ci/workflows/upgrade-stable-master.yaml
```diff
@@ -134,6 +134,10 @@ tasks:
     - "{{ .env.ARTIFACTS }}"
   force: true
   timeout: 5m
+  # kind export log is know to be flaky, so we are temporary ignoring errors in order
+  # to make the test past in case everything else passed
```
neolit123: minor typo past -> pass

kinder/pkg/test/workflow/taskCmdRunner.go
```diff
@@ -141,8 +141,8 @@ func (c *taskCmdRunner) Run(t *taskCmd, artifacts string, verbose bool) error {
 	// - the timeout is reached
 	select {
 	case err := <-result:
-		// if the command completed without errors, record the test case success and exits
-		if err == nil {
+		// if the command completed without errors or error should be ignored, record the test case success and exits
```
neolit123: > // if the command completed without an error or if we are ignoring errors, record the test case success and exit

fabriziopandini:
@neolit123 example of this working
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-kubeadm-kinder-upgrade-stable-master/1122285517438717956
 if if you search for task-11 in the build log, you can see that export actually log failed, but the error is ignored

Issue #1547, opened 2019-04-29
area/kubeadm label does not exist


fejta:
/assign @timothysc @fabriziopandini @neolit123

```yaml
 jsonPayload: {
  author:  "fabriziopandini"   
  component:  "hook"   
  event-GUID:  "ec05b660-6aab-11e9-9277-dc8d94bf201a"   
  event-type:  "pull_request"   
  level:  "warning"   
  msg:  "Unable to add nonexistent labels: ["area/kubeadm"]"   
  org:  "kubernetes"   
  plugin:  "owners-label"   
  pr:  1545   
  repo:  "kubeadm"   
  url:  "https://github.com/kubernetes/kubeadm/pull/1545"   
 }
```

Please remove this from the OWNERS file in this repo: https://github.com/kubernetes/kubeadm/blob/fc185e23ba206ac3b84befd6ff33e7c94f52235d/OWNERS#L25

Or create this label: https://github.com/kubernetes/test-infra/tree/master/label_sync

neolit123:
thanks for the heads up @fejta 
i will remove the label from the owners file as it's not needed in this repo.

PR #1548, opened 2019-04-29
OWNERS: remove redundant area/kubeadm label

neolit123:
The label is not needed for this repository as the
repository implies that this PR or ticket is already related
to kubeadm.

Having the label causes errors in the "owners-label" plugin:
fixes https://github.com/kubernetes/kubeadm/issues/1547

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -19,8 +19,7 @@ reviewers:
 - neolit123
 - rosti
 - yagonobre
-- ereslibre 
+- ereslibre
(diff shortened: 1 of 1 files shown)


OWNERS
```diff
@@ -19,8 +19,7 @@ reviewers:
 - neolit123
 - rosti
 - yagonobre
-- ereslibre 
+- ereslibre
```
neolit123: whitespace cleanup.

PR #1550, opened 2019-05-01
kinder: more presubmit test

fabriziopandini:
This PR adds more presubmit test (the kinder part of those test), so it will be easier to detect regressions on the kinder codebase

There are some TODO (further tests) to be addressed in following PR

/assign @neolit123


kinder/ci/kinder-presubmit-tests.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Kinder feature checks
+
+`pull-kubeadm-kinder-init-versions` job checks that kinder can create cluster for the supported init versions:
+
+- current -3/minor (e.g. v1.11)
+- current -2/minor (e.g. v1.12)
+- current -1/minor (e.g. v1.13)
+- current          (e.g. v1.14)
+- master           (e.g. v1.15)
```
neolit123: can we get away by only running the `master` one?
having runs on all the versions can delay PR merging quite a bit.
- we are already going to have coverage for these in periodics
- we can debug locally.

kinder/ci/workflows/variants.yaml
```diff
@@ -0,0 +1,183 @@
+version: 1
+summary: |
+  This workflow implements a sequence of tasks used test that kinder can manage supported kubeadm config variants.
```
neolit123: `used to test`

kinder/ci/kinder-presubmit-tests.md
neolit123: i'm not sure how fast it would be in CI, though.
with my internet it takes around 15 minutes to pull init + upgrade artifacts for the upgrade-stable-master job + a few more minutes to setup the cluster.

neolit123: something else we can do is have everything in the job, but only run in manually - e.g. `/test pull-kubeadm-kinder....`

fabriziopandini: I'm open to any option, but on CI it takes
- 1.5 minutes average for pulling 2 kubernetes version (here we need only 1)
- 2 minutes average for init
Overall It should take less than 20 minutes

neolit123: ok let's try `always_run` as a start, but possibly amend if it ends up being too slow.

Issue #1567, opened 2019-05-15
Tracking issue for HA to beta


fabriziopandini:
What is missing for getting HA to beta in v1.15?

- [x] “--experimental-control-plane” flag to be renamed “--control-plane”
- [x] “--experimental-upload-certs” flag to be renamed “--upload-certs”. Default to true?
- [x] https://github.com/kubernetes/kubeadm/issues/1071 “kubeadm upgrade node experimental-control-plane” to be merged into “kubeadm upgrade node”
- [punted to v1.16] https://github.com/kubernetes/kubeadm/issues/1432 make` join --control-plane` more resilient (it should work after loosing one control plane/etcd) https://github.com/kubernetes/kubernetes/pull/75600
- [x] e2e tests
   - [x] change kind-master job (and possibly kind-1.14) to test HA
   - [x] change upgrade-stable-master job (and possibly  upgrade-1.13-1.4) to test HA + automatic copy certificates
   - [x]  change master-on-stable job (and possibly 1.14 on 1.13) to test HA + automatic copy certificates
   - [x]  create a new job testing HA + automatic copy certificates with external etcd

fabriziopandini:
@neolit123 opinions about the E2E part? the main idea behind is to not recreate a new series of job for HA, but instead, make current jobs have an additional control plane node so they are testing also HA

neolit123:
>  “--experimental-upload-certs” flag to be renamed “--upload-certs”. Default to true?

needs discussion with the group.

> “kubeadm upgrade node experimental-control-plane” to be merged into “kubeadm upgrade node”

i'd be interested in how the detection will work here.

> @neolit123 opinions about the E2E part? the main idea behind is to not recreate a new series of job for HA, but instead, make current jobs have an additional control plane node so they are testing also HA

+1 on switching everything to HA.

> create a new job testing HA + automatic copy certificates with external etcd

seems good. i think we can get away without upgrade for external etcd for the beta...

neolit123:
> +1 on switching everything to HA.

also i can work on these if you'd like.

RA489:
@fabriziopandini i can work on some of them.
 “--experimental-control-plane” flag to be renamed “--control-plane”
  “--experimental-upload-certs” flag to be renamed “--upload-certs”. Default to true?

neolit123:
hi @RA489 we spoke about these during yesterday's kubeadm office hours.
the idea was to do these PR last from the list from the list above.

[...]

PR #1580, opened 2019-05-29
kinder make upgrade and skew e2e tests HA ready

fabriziopandini:
This PR:
- change upgrade-stable-master job and upgrade-1.13-1.4 to test HA + automatic copy certificates
- change master-on-stable job and 1.14 on 1.13 to test HA + automatic copy certificates

rif https://github.com/kubernetes/kubeadm/issues/1567

NB. In case I'm out of sync with other helping to this effort, feel free to close.

/kind test
/assign @neolit123 

### DIFF
--- a/kinder/ci/workflows/skew-1.14-on-1.13.yaml
+++ b/kinder/ci/workflows/skew-1.14-on-1.13.yaml
@@ -7,5 +7,6 @@ summary: |
 vars:
   kubeadmVersion: "{{ resolve `ci/latest-1.14` }}"
   kubernetesVersion: "{{ resolve `ci/latest-1.13` }}"
+  controlPlaneNodes: 2
 tasks:
 - import: skew-x-on-y-tasks.yaml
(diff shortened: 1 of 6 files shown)


kinder/ci/workflows/skew-1.14-on-1.13.yaml
```diff
@@ -7,5 +7,6 @@ summary: |
 vars:
   kubeadmVersion: "{{ resolve `ci/latest-1.14` }}"
   kubernetesVersion: "{{ resolve `ci/latest-1.13` }}"
+  controlPlaneNodes: 2
```
neolit123: it think it be 3 instead of 2 for HA as per our docs and the redundancy factor.

fabriziopandini: @neolit123 considering that the goal of this test is to check `join --control-plane` & `upgrade node`, (not the etcd clustering), two is enough.
However, no problem to change it

neolit123: i've seen instances where users following our guide fail on the 3rd control plane node.
(of course due to a user mistake, for one reason or another).

/lgtm

neolit123:
missed to add `/hold`
@fabriziopandini if we are adding 2 control plane nodes instead of 3 then the tests are no longer HA, but only "multi-control" plane tests.

fabriziopandini:
@neolit123 point taken
I will send a patch ASAP

PR #1590, opened 2019-06-01
kinder: v1beta2 workaround

fabriziopandini:
This PR implements a workaround for using kinder to test v1beta2 kubeadm config version, even if kind does not support this yet.

It also adapts the init workflow to pass the `certificate-key` using the config file instead of the flag

/kind feature
/assign @neolit123
/assign @rosti


ereslibre:
Any reason to not adapt https://github.com/kubernetes/kubeadm/blob/523d4bdcdc7598323f43c78bb9977db44f1e59c8/kinder/pkg/actions/kubeadm-join.go#L168-L172 as well?

fabriziopandini:
@ereslibre currently kubeadm-join is not using `--config`, but flags
This can be eventually fixed in a follow-up PR

PR #1605, opened 2019-06-12
fix-upgrade-flake

fabriziopandini:
This PR introduces a workaround for reducing flakes from the upgrade test.
However, the real source of the problems still is not clear and should be investigated

/kind flake
/assign @neolit123

### DIFF
--- a/kinder/pkg/actions/waiter.go
+++ b/kinder/pkg/actions/waiter.go
@@ -231,11 +231,20 @@ func staticPodHasVersion(pod, version string) func(kctx *kcluster.KContext, kn *
 	}
 }
 
-// kubeletHasRBAC is a test checking that kubelet has access to the expected config map
+// kubeletHasRBAC is a test checking that kubelet has reliable access to kubelet-config-x.y and kube-proxy,
+// where reliable = it have access for 5 seconds in a row.
+//
+// this test is a workaround meant to prevent errors like: configmaps "kube-proxy" is forbidden:
+// User "system:node:kinder-upgrade-control-plane3" cannot get resource "configmaps" in API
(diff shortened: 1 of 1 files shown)


kinder/pkg/actions/waiter.go
```diff
@@ -231,19 +231,36 @@ func staticPodHasVersion(pod, version string) func(kctx *kcluster.KContext, kn *
...
+				"cm",
+				"kube-proxy",
+				"--kubeconfig=/etc/kubernetes/kubelet.conf",
+				"-n=kube-system",
+				"-o=jsonpath='{.metadata.name}'",
+			)
+			if output1 != "" && output2 != "" {
 				time.Sleep(1 * time.Second)
 				continue
 			}
```
neolit123: if one of the "get cm" fails in any iteration of the loop the function will return false right away.
i wonder if the problem we are seeing should be solved with retry-on-fail (e.g. 5 times in a row) instead of checking reliability?

fabriziopandini: there is a bit of layering here
- There is a waiter cycle that waits for a condition to be satisfied up to 5 minutes
- The internal condition check that you can read both the CM for 5 times in a row

> if one of the "get cm" fails in any iteration of the loop the function will return false right away

No the condition will be retried by the external wait cycle up to 5 minutes

> i wonder if the problem we are seeing should be solved with retry-on-fail (e.g. 5 times in a row) instead of checking reliability?

This requires much more thinking, because kubeadm does not return errors codes, so you can't understand which type of error are you facing

in a few words, this is a workaround, far for being perfect, but it should reduce flakes while we investigate the real source of problems

neolit123: ok, i didn't check this external wait cycle.
/lgtm

PR #1636, opened 2019-06-26
kinder: avoid polution of the output of "kubeadm config images list"

neolit123:
In etcd.go when we try to get the propper version of etcd
to use from "kubeadm config images list", if the version
of kubeadm is much older than "stable-1" (which is the default label
in kubeadm), the ouput shows a warning to stderr e.g.:

```
  I0626 02:54:47.931193    3673 version.go:240] remote version
  is much newer: v1.15.0; falling back to: stable-1.14
```

For some reason this output becomes part of the "etcdVersionBuf"
which is supposed to only contain stdout (cmd.SetStdout(&etcdVersionBuf)).

Omit this ouput by piping stderr to /dev/null and only
catch the etcd image string - e.g.:

```
  k8s.gcr.io/etcd:3.3.10
```

Once kubeadm supports machine readable output we can parse
this output properly without workarounds.

-----
fixes failures for the external etcd 1.14 test:
https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-kubeadm#kubeadm-kinder-external-etcd-1.14

second commit enables debug output for the kinder external etcd worfkflows.

/assign @ereslibre 
cc @fabriziopandini 
/kind failing-test
/priority important-soon

### DIFF
--- a/kinder/ci/workflows/external-etcd.yaml
+++ b/kinder/ci/workflows/external-etcd.yaml
@@ -40,6 +40,7 @@ tasks:
     - --control-plane-nodes=3
     - --worker-nodes=2
     - --external-etcd
+    - --loglevel=debug
   timeout: 5m
 - name: init
   description: |
@@ -51,6 +52,7 @@ tasks:
     - kubeadm-init
     - --automatic-copy-certs
     - --name={{ .vars.clusterName }}
+    - --loglevel=debug
   timeout: 5m
 - name: join
   description: |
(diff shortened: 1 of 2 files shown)


ereslibre:
Thanks @neolit123!

Wondering about:

> For some reason this output becomes part of the "etcdVersionBuf" which is supposed to only contain stdout (cmd.SetStdout(&etcdVersionBuf)).

However I'm fine with this change, but we might want to check why this is happening in any case if we find the time...

/lgtm

neolit123:
> However I'm fine with this change, but we might want to check why this is happening in any case if we find the time...

logged https://github.com/kubernetes-sigs/kind/issues/659

PR #1673, opened 2019-07-17
Fix image preload

fabriziopandini:
This PR fixes the E2E tests failing after https://github.com/kubernetes/kubernetes/pull/80047 merged

The PR has two commits:
- `fix-image-tar that` introduces the actual fix, adjusting the image name with `-amd64` suffix during the alter image process
- `add check prepulled images` that introduces a new check before main kubeadm commands; the check compares the images kubeadm is going to use with the images already existing on the node and reports what is missing. In normal conditions, the output should be something similar to:
```
kinder-upgrade-control-plane2:$ Checking pre-loaded images
Some of the required images are not pre-loaded into the container runtime:
k8s.gcr.io/pause:3.1
k8s.gcr.io/etcd:3.3.10
k8s.gcr.io/coredns:1.3.1
```
if also the control plane images are missing, this is a signal something is wrong...

/assign @neolit123

### DIFF
--- a/kinder/pkg/build/alter/alter.go
+++ b/kinder/pkg/build/alter/alter.go
@@ -18,7 +18,9 @@ package alter
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
+	"strings"
 
 	"github.com/google/uuid"
 	"github.com/pkg/errors"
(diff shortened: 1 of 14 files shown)


kinder/pkg/build/alter/alter.go
```diff
@@ -185,13 +188,80 @@ func (c *Context) prepareBits(bitsInstallers []bits.Installer, bc *bits.BuildCon
...
+	if strings.HasSuffix(repository, archSuffix) {
+		fixed := strings.TrimSuffix(repository, archSuffix)
+		fmt.Println("fixed: " + repository + " -> " + fixed)
+		repository = fixed
+	}
+
+	return repository
+}
+
+func contains(s []string, e string) bool {
```
neolit123: possibly better to move this is a closure under `prepareBits` if it's used only there, or leave it package wider but rename it to something more descriptive - e.g. `stringSliceContainsString`

fabriziopandini:
@neolit123 comment addressed!

neolit123:
thanks
/lgtm

PR #1702, opened 2019-08-01
Kinder: Test file discovery

fabriziopandini:
This PR:
- implements required changes in order to use kinder for testing `kubeadm join` discovery modes
- implements and E2E test workflow for testing `kubeadm join` discovery modes
- updates documents with latest changes
NB. commits are organized according to ^^^

Note for reviewers:
- In order to make this test possible, it was necessary to improve the implementation of the support for executing actions on a single node (rif only-node/skip)
- While testing new config combinations, an error on the docker patch was fixed


kinder/pkg/cluster/manager/actions/kubeadm-config.go
```diff
@@ -272,14 +325,101 @@ func getKubeadmConfig(c *status.Cluster, n *status.Node, data kubeadm.ConfigData
 	// if the node is the bootstrap control plane, then all the objects used as init time
 	if n == c.BootstrapControlPlane() {
 		return selectYamlFramentByKind(patched,
+			"MasterConfiguration", //v1alpha2
```
neolit123: v1alpha2 was removed in 1.13, while v1alpha3 was already available in 1.12.
1.12 is outside of the skew but we must support it for upgrades.
i think we can avoid the v1alpha2 MasterConfiguration and NodeConfiguration objects and use v1alpha3, but i may be missing something?

kinder/pkg/cluster/manager/actions/kubeadm-config.go
```diff
@@ -272,14 +325,101 @@ func getKubeadmConfig(c *status.Cluster, n *status.Node, data kubeadm.ConfigData
...
+		"NodeConfiguration", //v1alpha2
+		"JoinConfiguration",
+	), nil
+}
+
+func createDiscoveryFile(c *status.Cluster, n *status.Node, discoveryMode DiscoveryMode) error {
+	// the discovery file is a kubeaconfig file, so for sake of semplicity in setting up this test,
+	// we are using the admin.conf file created by kubeadm on the bootstrap control plane node
+	// as a starting point (e.g. it already contains the necessary server address/server certificate)
+	// IMPORTANT. Don't do this in production, admin.conf contains cluster-admin credentials.
```
neolit123: SGTM for testing, i don't think we need to test temporary (soon to expire) credentials for now.

kinder/pkg/cluster/manager/actions/kubeadm-config.go
```diff
@@ -272,14 +325,101 @@ func getKubeadmConfig(c *status.Cluster, n *status.Node, data kubeadm.ConfigData
...
+	lines, err := c.BootstrapControlPlane().Command(
+		"cat", "/etc/kubernetes/admin.conf",
+	).Silent().RunAndCapture()
+	if err != nil {
```
neolit123: can be:
`if err != nil || len(lines) == 0 {`

kinder/pkg/constants/constants.go
```diff
@@ -89,6 +89,9 @@ const (
 	// CertificateKey defines a dummy, well known CertificateKey for automating automatic copy certs process
 	// const CertificateKey = "d02db674b27811f4508bf8a5fa19fbe060921340552f13c15c9feb05aaa96824"
 	CertificateKey = "0123456789012345678901234567890123456789012345678901234567890123"
+
+	// DiscoveryFile defines the path to a discovery file stored on nodes
+	DiscoveryFile = "/kinder/discovery.yaml"
```
neolit123: optional: use `.conf`?
i don't think kubeconfig files have it as a standard extension, though.

kinder/ci/workflows/discovery-master.yaml
```diff
@@ -0,0 +1,147 @@
+version: 1
+summary: |
+  This workflow implements a sequence of tasks used for testing alternative
+  discovery methods for kubeadm join.
+vars:
+  kubernetesVersion: tmp/v1.16.0-alpha.2.128+86ebaaa0232999 #"{{ resolve `ci/latest` }}"
```
neolit123: why not use the "ci/latest" label; the latest fix changes are already in master?

kinder/pkg/cluster/manager/actions/kubeadm-config.go
fabriziopandini: you are right, we can nuke v1alpha2, but this is out of the scope of this PR

kinder/pkg/cluster/manager/actions/kubeadm-config.go
fabriziopandini: if `err==nil` and `len(lines) == 0`, which happen e.g. if you try to create discovery file before init, then you call `errors.Wrapf(err, ...` with `err==nil` which sadly returns `nil` instead of an error. 
So we have to break conditions in two

fabriziopandini:
@neolit123 thanks! some comments addressed + answers ^^^

PR #1728, opened 2019-08-16
fix misspelled words in runner.go

ethan-daocloud:
Signed-off-by: Guangming Wang <guangming.wang@daocloud.io>


neolit123:
thanks for the fix @ethan-daocloud

PR #1737, opened 2019-08-21
kinder: add hack/verify-staticcheck.sh

neolit123:
Enable static checks for .go files which was recently enabled in
the kubernetes/kubernetes repository too.

/kind feature cleanup
/priority backlog


dixudx:
@neolit123 So the changes here are cherry-picked from k/k ?

neolit123:
@dixudx 
no, this is a new implementation of the shell script.
k/k skips a lot of checks so this is more in the lines of what sigs.k8s.io/kind is doing (running all checks).

neolit123:
i will send a second commit that fixes the discovered staticcheck problems.

neolit123:
@dixudx @fabriziopandini updated, PTAL.

Issue #1695, opened 2019-07-29
E2E Testing Documentation has been invalidated by Testgrid Updates


chases2:
The documentation on [managing end-to-end tests](https://github.com/kubernetes/kubeadm/blob/master/docs/managing-e2e-tests.md) in this repository has grown stale as Testgrid has been enhanced.

Specifically:
- "config.yaml" no longer exists; testgrid configurations live in their [own directory](https://github.com/kubernetes/test-infra/tree/master/config/testgrids). Those links are broken.
- Many of the jobs referenced, such as the example file [kubeadm.yaml](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-cluster-lifecycle/kubeadm-kind.yaml), now use prow annotations instead of only a YAML Testgrid configuration
- The Testgrid configs pertinent to this doc and repository [should](https://github.com/kubernetes/test-infra/issues/13541) be separated into their own file. If this is done, then the document could be simplified since grepping through a large YAML file isn't needed anymore.

neolit123:
thanks for the reminder @chases2

PR #1746, opened 2019-08-23
managing-e2e-tests: refactor the document

neolit123:
- Provide some instructions for updating kinder workflows.
- Reorganize the sections a little: introduction, configurations,
how-to do updates, more info
- Add information on using prow-job annotations and remove
the old way of using testgrid config entries.

fixes https://github.com/kubernetes/kubeadm/issues/1695

### DIFF
--- a/docs/managing-e2e-tests.md
+++ b/docs/managing-e2e-tests.md
@@ -6,12 +6,13 @@ Kubernetes has a rich end-to-end (e2e) testing infrastructure, which allows deta
 
 Kubernetes uses applications such as the web-based [testgrid](https://k8s-testgrid.appspot.com/) for monitoring the status of e2e tests. test-infra also hosts the configuration on individual test jobs and Docker images that contain tools to invoke the jobs.
 
-### Details on kubeadm test jobs
+### Prow job configuration
 
 The following folder contains all the SIG Cluster Lifecycle (the SIG that maintains kubeadm) originated test jobs:
 [sig-cluster-lifecycle](https://github.com/kubernetes/test-infra/tree/master/config/jobs/kubernetes/sig-cluster-lifecycle)
 
-Please note that this document will only cover details on the `kubeadm*.yaml` files.
+Please note that this document will only cover details on the `kubeadm*.yaml` files and only on some of the parameters
+these files contain.
 
 For example, let's have a look at this file:
 [kubeadm.yaml](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-cluster-lifecycle/kubeadm-kind.yaml)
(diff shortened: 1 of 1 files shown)


docs/managing-e2e-tests.md
```diff
@@ -23,106 +24,140 @@ It contains a list of jobs such as:
...
+In this case, `ci-kubernetes-e2e-kubeadm-kind-master` is a test job that runs every 2 hours and it also
+has a set of other parameters defined for it. Such test jobs use an image that will run as a container inside
+a Pod of the Kubernetes [Prow](https://github.com/kubernetes/test-infra/tree/master/prow) cluster.
 
-Important to note here is that testgrid jobs have a docker `image` that contains all the tools needed to run tests. Among them, one of the most relevant ones, is the tool that will be used for the deployment of the cluster.
+For this example job the deployment tool is called [kind](https://github.com/kubernetes-sigs/kind) (with integration
+in [this file](https://github.com/kubernetes/test-infra/blob/master/kubetest/kind/kind.go)) and the image is called
+`*kubekins:...`. As a very high level summary, the way this works is when a job is invoked all the job parameters
+are passed to a CLI tool called kubetest, which then instantiates the job container and then the deployment tool
+inside it. Please note that kubetest is not required an test authors can decide to call any bash script instead.
```
fabriziopandini: and test authors

docs/managing-e2e-tests.md
```diff
@@ -23,106 +24,140 @@ It contains a list of jobs such as:
   ...
 ```
 
-In this case, `ci-kubernetes-e2e-kubeadm-kind-master` is a test job that runs every 2 hours and it also has a set of parameters defined for it. This document will only cover some of them.
+In this case, `ci-kubernetes-e2e-kubeadm-kind-master` is a test job that runs every 2 hours and it also
+has a set of other parameters defined for it. Such test jobs use an image that will run as a container inside
+a Pod of the Kubernetes [Prow](https://github.com/kubernetes/test-infra/tree/master/prow) cluster.
 
-Important to note here is that testgrid jobs have a docker `image` that contains all the tools needed to run tests. Among them, one of the most relevant ones, is the tool that will be used for the deployment of the cluster.
+For this example job the deployment tool is called [kind](https://github.com/kubernetes-sigs/kind) (with integration
```
fabriziopandini: I'm not sure all those detail are useful/easy to understand for a broader audience.
what about leaving only the high level summary?

docs/managing-e2e-tests.md
```diff
@@ -23,106 +24,140 @@ It contains a list of jobs such as:
...
-    test_group_name: ci-kubernetes-e2e-kubeadm-kind-...
-```
+As an overview:
+- SIG Cluster Lifecycle dashboards reside in a dashboard group that can be found [here](https://k8s-testgrid.appspot.com).
+- Inside this dashboard group there is a [dashboard for kubeadm](https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-kubeadm).
+- Inside this dashboard there are individual tabs such as `kubeadm-kind-master` (which is the tab name for the
+job `ci-kubernetes-e2e-kubeadm-kind-master`).
 
-What the above does is defining a dashboard called `sig-cluster-lifecycle-all` and placing inside a couple of tests defined by `name` and linking to a `test_group_name`. Some items may also have a `description`.
+In earlier versions of testgrid, all the different test jobs of different SIGs were contained in single
```
fabriziopandini: The history is informative, but probably it is better to keep things simple and describe only the current state. wdyt?

docs/managing-e2e-tests.md
```diff
@@ -23,106 +24,140 @@ It contains a list of jobs such as:
...
+have to be removed and `1.15` jobs have to be added. The project would support version `15 - 2 = 13` as the minimum.
 
-#### Updating test jobs
+#### Updating kinder workflows
+
+Kinder is deployment tool that currently resides in the kubernetes/kubeadm repository.
+It contains a folder with workflows that need update each cycle:
+https://github.com/kubernetes/kubeadm/blob/master/kinder/ci/workflows/
+
+The summary of the actions is the following:
```
fabriziopandini: before the list, might be worth to mention the assumptions: 
if Kubernetes 1.15 is about to be released soon (and 1.12 version would go outside of the support),

docs/managing-e2e-tests.md
```diff
@@ -23,106 +24,140 @@ It contains a list of jobs such as:
...
+- If there are `1.12 -> 1.13` upgrade jobs leave them, because upgrades to `1.13` still have to be supported.
+- Remove `1.13 on 1.12` jobs.
```
fabriziopandini: skew jobs

fabriziopandini: Additionally, if we look at this cycle, this set of actions is going to be executed in two phases:
1. create new jobs (when the release branch is created)
2. delete old jobs (after the release is cut)
What about describing this accordingly?

docs/managing-e2e-tests.md
```diff
@@ -23,106 +24,140 @@ It contains a list of jobs such as:
...
+- Add upgrade jobs for `1.14 -> 1.15` and `1.15 on 1.14` skew jobs.
+- Add plain `1.15` jobs.
+- Make sure that workflows include the correct `ci/latest-x` labels.
+- Make sure that workflows have the correct testgrid URLs.
 
-This document will cover information on how to perform updates on these 3 files:
-- [kubeadm-kind.yaml](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-cluster-lifecycle/kubeadm-kind.yaml "kubeadm-kind.yaml")
+#### Updating test jobs in kubernetes/test-infra
+
+This document will cover information on how to perform updates on these files:
```
fabriziopandini: I think we should link to  https://github.com/kubernetes/kubeadm/blob/master/docs/release-cycle.md instead of maintaining another list of test

docs/managing-e2e-tests.md
```diff
@@ -143,7 +178,7 @@ Note that test-infra requires [Bazel](https://bazel.build/).
 
 At this point you can commit your changes and send a GitHub PR against the test-infra repository.
 
-#### Job run frequency
+### Job run frequency
```
fabriziopandini: I think that this should be documented in  https://github.com/kubernetes/kubeadm/blob/master/docs/release-cycle.md as well

docs/managing-e2e-tests.md
```diff
@@ -23,106 +24,140 @@ It contains a list of jobs such as:
...
 
+- [kubeadm-kinder-external-etcd.yaml](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-cluster-lifecycle/kubeadm-kinder-external-etcd.yaml)
+
+Holds tests that deploy a kubeadm / kinder cluster using external etcd, instead of the default stacked etcd.
+
 The files can be found in the [sig-cluster-lifecycle](https://github.com/kubernetes/test-infra/tree/master/config/jobs/kubernetes/sig-cluster-lifecycle) folder.
 
-The idea is to remove old test jobs and add new ones. If Kubernetes 1.15 is about to be released soon, this means that the `15 - 3 = 12` MINOR version would go outside of the support skew of the project. `1.12` jobs have to be removed and `1.15` jobs have to be added. The project would support version `15 - 2 = 13` as the minimum.
 
+The summary of the actions is the following:
```
fabriziopandini: This list, at least at the begging, is the same as the one above under "Updating kinder workflows".
What about leaving only recommendations for creating new jobs?

docs/managing-e2e-tests.md
```diff
@@ -23,106 +24,140 @@ It contains a list of jobs such as:
...
+
+Ideally both PRs should be merged around the same time, so that there are no failed runs.
 
-There is work in progress in test-infra on tools that can hopefully automate this process.
+The idea is to remove old tests and add new ones. For example. if Kubernetes 1.15 is about to be released soon,
+this means that the `15 - 3 = 12` MINOR version would go outside of the support skew of the project. `1.12` jobs
+have to be removed and `1.15` jobs have to be added. The project would support version `15 - 2 = 13` as the minimum.
 
-#### Updating test jobs
+#### Updating kinder workflows
```
fabriziopandini: Is this kinder specific?
are you palling to write something similar for kind or some of the considerations apply to both...

docs/managing-e2e-tests.md
```diff
@@ -23,106 +24,140 @@ It contains a list of jobs such as:
...
+
+Kinder is deployment tool that currently resides in the kubernetes/kubeadm repository.
```
fabriziopandini: what about simplifying: Kinder workflows need to be updated each cycle

docs/managing-e2e-tests.md
neolit123: removed the kind.go integration links.

docs/managing-e2e-tests.md
neolit123: removing the historic info

docs/managing-e2e-tests.md
neolit123: the assumptions are in the previous section `Updates to kubeadm tests` and they apply to both updating test-infra and kinder workflows.

docs/managing-e2e-tests.md
neolit123: updated

docs/managing-e2e-tests.md
neolit123: https://github.com/kubernetes/kubeadm/blob/master/docs/release-cycle.md is outdated.
i think instead we should link from https://github.com/kubernetes/kubeadm/blob/master/docs/release-cycle.md to this document that is specific for tests.

docs/managing-e2e-tests.md
neolit123: https://github.com/kubernetes/kubeadm/pull/1746#discussion_r317841591

docs/managing-e2e-tests.md
neolit123: consolidated the list of actions:
```
### Updates to kubeadm tests
...
The summary of the actions is the following:
- shared actions here

#### Updating kinder workflows
- specific actions to kinder

#### Updating test jobs in kubernetes/test-infra
- specific actions to test infra

```

docs/managing-e2e-tests.md
neolit123: for kind we don't need one, because it's managed by updating test-infra job YAML.
for kinder on the other hand we have to touch workflow files.

neolit123:
thanks for the comments @fabriziopandini 
updated  + added some responses.

fabriziopandini:
@neolit123 
Do you think we can unhold this? it is definitely a step forward and probably we can do another iteration when we look at https://github.com/kubernetes/kubeadm/blob/master/docs/release-cycle.md  for transforming it into a checklist

neolit123:
@fabriziopandini 
SGTM
/hold cancel

agreed, the release cycle doc needs a major update + possibly include a template that can be copy pasted into a tracking issue each cycle (as we discussed).

PR #1750, opened 2019-08-23
Move kinder to kind v0.5.1, update deps

stealthybox:
/area kinder
/area dependency
/kind feature

Packages were moved.
The PatchJSON struct moved.
I made sure all of the third_party forks' individual files still matched the current ones from kind.

This seems to build and work well for me.
There don't appear to be many runnable tests.

We may not merge this before 1.16 since the current kinder is critical to test-grid.

@fabriziopandini @neolit123


kinder/pkg/cluster/manager/actions/kubeadm-config.go
```diff
@@ -29,7 +29,8 @@ import (
 	"k8s.io/kubeadm/kinder/pkg/constants"
 	"k8s.io/kubeadm/kinder/pkg/cri"
 	"k8s.io/kubeadm/kinder/pkg/kubeadm"
-	kindkustomize "sigs.k8s.io/kind/pkg/kustomize"
+	kindkustomize "k8s.io/kubeadm/kinder/third_party/kind/util/kustomize"
+	"sigs.k8s.io/kind/pkg/apis/config/v1alpha3"
```
fabriziopandini: [1] We are always adding aliases for kind imports. This can be 
`kindV1alpha3 "sigs.k8s.io/kind/pkg/apis/config/v1alpha3"`

kinder/pkg/cluster/status/cluster.go
```diff
@@ -24,8 +24,8 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	"k8s.io/kubeadm/kinder/pkg/constants"
+	"sigs.k8s.io/kind/pkg/apis/config/v1alpha3"
```
fabriziopandini: [1]

kinder/pkg/config/config.go
```diff
@@ -21,7 +21,7 @@ import (
 	"os"
 
 	"github.com/pkg/errors"
-	"sigs.k8s.io/kind/pkg/cluster/config/v1alpha3"
+	"sigs.k8s.io/kind/pkg/apis/config/v1alpha3"
```
fabriziopandini: [1]

kinder/pkg/kubeadm/discovery.go
```diff
@@ -24,16 +24,16 @@ import (
 
 	K8sVersion "k8s.io/apimachinery/pkg/util/version"
 	"k8s.io/kubeadm/kinder/pkg/constants"
-	kindkustomize "sigs.k8s.io/kind/pkg/kustomize"
+	"sigs.k8s.io/kind/pkg/apis/config/v1alpha3"
```
fabriziopandini: [1]

kinder/pkg/cluster/manager/actions/kubeadm-config.go
```diff
@@ -29,7 +29,8 @@ import (
 	"k8s.io/kubeadm/kinder/pkg/constants"
 	"k8s.io/kubeadm/kinder/pkg/cri"
 	"k8s.io/kubeadm/kinder/pkg/kubeadm"
-	kindkustomize "sigs.k8s.io/kind/pkg/kustomize"
+	kindkustomize "k8s.io/kubeadm/kinder/third_party/kind/util/kustomize"
```
fabriziopandini: This should be `kindinternalkustomize` (we are using `kindinternal` prefix for everything we are forking)

kinder/pkg/cluster/manager/actions/kubeadm-config.go
neolit123: my idea with this file was to not import kind versions directly and it as a wrapper:
https://github.com/kubernetes/kubeadm/blob/master/kinder/pkg/config/config.go

currently it's the only place where kind's v1alpha3 is imported, which makes changes in kinder public types easier to manage in kinder.

neolit123: aliasing v1alpha3.PatchJSON6902 in there should be sufficient.

neolit123: isn't the only nesting here: https://github.com/kubernetes-sigs/kind/blob/master/pkg/apis/config/v1alpha3/types.go#L51
?

neolit123:
odd, upstream kind is already running in CI and their test jobs don't have these failures:
https://k8s-testgrid.appspot.com/sig-testing-kind

neolit123:
on a quick look these seem unrelated to kinder upgrading the cluster. :thinking:

neolit123:
> The Node image the workflow is building for the test-suite is using an incompatible base:

yeah, that must be it. we must bump workflows too.

btw, i'm hesitant about updating this before the 1.16 release, reason is that we are making updates to etcd, coredns and we need to have good signal. but will defer to @fabriziopandini on the final call.

fabriziopandini:
@stealthybox thanks for addressing all the feedbacks!

I'm ok in waiting till after 1.16 because bumping the image is tricky, because this will shift all our tests from docker to containerd and we don't want this to happen
I'm opening an issue to track this.

fabriziopandini:
If test with `kindest/base:v20190403-1ebf15f` are failing due to https://github.com/kubernetes-sigs/kind/issues/771, might be this can be fixed by adding a volume here https://github.com/kubernetes/kubeadm/blob/45a788e30f7ab46f3adcf5286ec0e371b92ff385/kinder/pkg/cri/docker/createhelper.go#L94

fabriziopandini:
@stealthybox 
see https://github.com/kubernetes/kubeadm/issues/1759
Kinder now can build docker images https://github.com/kubernetes/kubeadm/pull/1768
have you had the chance to test the change suggested [here](https://github.com/kubernetes/kubeadm/pull/1750#issuecomment-526215134)

fabriziopandini:
@stealthybox have you got some time to check my last comments?

neolit123:
just a note that kind has seen a lot of changes after 0.5.1 too, so we should try to keep up.

fabriziopandini:
@stealthybox this is getting urgent...
If you don't have time I can pick up this effort

stealthybox:
Just noting here that I promised to pick this up to support our upcoming release.

PR #1760, opened 2019-08-29
Set calico `rp_filter` sysctl for all nodes (including load-balancer)

stealthybox:
/kind bug
/area kinder

After performing `kinder do kubeadm-init` on a cluster with worker nodes,
the calico-node pods on the worker nodes will CrashLoop.

This is because this `rp_filter` sysctl is only set on the control-plane nodes.
I'm changing this to all nodes, because there is no compelling reason for any external-load-balancer or other cluster-infrastructure to have a different kernel-routing setup than the k8s nodes.

I'd also be fine to move this to the end of `createNodes()` if that is preferred.

/assign @fabriziopandini
/assign @neolit123


stealthybox:
/cc @BenTheElder
This patch should improve our ci-signal

PR #1761, opened 2019-08-30
verify_manifest_lists: skip cloud-controller-manager for 1.16

neolit123:
xref https://github.com/kubernetes/kubernetes/pull/81029/files

The CCM image was removed for the release of 1.16.
Skip checking this image for versions 1.16*.

Also use go modules for running the test using the shell script, which makes things much faster if e.g. once has the apimachinery module locally already.

### DIFF
--- a/tests/e2e/manifests/verify_manifest_lists.sh
+++ b/tests/e2e/manifests/verify_manifest_lists.sh
@@ -36,7 +36,7 @@ curl -sS "https://raw.githubusercontent.com/kubernetes/apimachinery/master/pkg/u
 popd
 
 # set GOPATH to the local directory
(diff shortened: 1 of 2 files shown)


tests/e2e/manifests/verify_manifest_lists.sh
```diff
@@ -36,7 +36,7 @@ curl -sS "https://raw.githubusercontent.com/kubernetes/apimachinery/master/pkg/u
 popd
 
 # set GOPATH to the local directory
```
andrewsykim: outdated comment now :P

neolit123:
well it merged too fast. :)

PR #1767, opened 2019-09-02
kinder: workaround RBAC flakiness  during upgrades

fabriziopandini:
This PR tries to improve the current workaround for RBAC flakiness during upgrades
/assign @neolit123

### DIFF
--- a/kinder/pkg/cluster/manager/actions/waiter.go
+++ b/kinder/pkg/cluster/manager/actions/waiter.go
@@ -291,23 +291,23 @@ func staticPodHasVersion(pod, version string) func(c *status.Cluster, n *status.
 func kubeletHasRBAC(major, minor uint) func(c *status.Cluster, n *status.Node) bool {
 	return func(c *status.Cluster, n *status.Node) bool {
 		for i := 0; i < 5; i++ {
-			output1 := kubectlOutput(c.BootstrapControlPlane(),
+			output1 := kubectlOutput(n,
+				"auth",
+				"can-i",
 				"get",
-				"cm",
-				fmt.Sprintf("kubelet-config-%d.%d", major, minor),
 				"--kubeconfig=/etc/kubernetes/kubelet.conf",
-				"-n=kube-system",
-				"-o=jsonpath='{.metadata.name}'",
+				"--namespace=kube-system",
+				fmt.Sprintf("configmaps/kubelet-config-%d.%d", major, minor),
 			)
-			output2 := kubectlOutput(c.BootstrapControlPlane(),
+			output2 := kubectlOutput(n,
+				"auth",
+				"can-i",
 				"get",
-				"cm",
-				"kube-proxy",
 				"--kubeconfig=/etc/kubernetes/kubelet.conf",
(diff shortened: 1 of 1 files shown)


kinder/pkg/cluster/manager/actions/waiter.go
```diff
@@ -291,14 +291,16 @@ func staticPodHasVersion(pod, version string) func(c *status.Cluster, n *status.
 func kubeletHasRBAC(major, minor uint) func(c *status.Cluster, n *status.Node) bool {
 	return func(c *status.Cluster, n *status.Node) bool {
 		for i := 0; i < 5; i++ {
```
neolit123: should we increase the number of retries?
i think even 30 seconds here are fine.

fabriziopandini: without any real clue on the underlying problem, I prefer to make the first try only improving the validation of the kubectl output...

BTW, this are not retries, but the number of consecutive times in a row the check should pass before assuming we are in a reliable state.

kinder/pkg/cluster/manager/actions/waiter.go
```diff
@@ -307,7 +309,7 @@ func kubeletHasRBAC(major, minor uint) func(c *status.Cluster, n *status.Node) b
 				"-n=kube-system",
 				"-o=jsonpath='{.metadata.name}'",
 			)
-			if output1 != "" && output2 != "" {
+			if strings.Trim(output1, "'") == name1 && strings.Trim(output2, "'") == name2 {
```
neolit123: i think we can use:
```
kubectl auth can-i get --kubeconfig /etc/kubernetes/kubelet.conf --namespace kube-system configmaps/kubelet-config-x.yy
kubectl auth can-i get --kubeconfig /etc/kubernetes/kubelet.conf --namespace kube-system configmaps/kube-proxy
```

```
$ sudo kubectl auth can-i get --kubeconfig /etc/kubernetes/kubelet.conf --namespace kube-system configmaps/kube-proxy
yes
```

kinder/pkg/cluster/manager/actions/waiter.go
neolit123: i think you already explained this to me, but i again forgot.
i guess what i'm confused about is why the state can suddenly become unreliable.

once RBAC is applied it should remain there :thinking:

kinder/pkg/cluster/manager/actions/waiter.go
fabriziopandini: do you know when this feature was introduced? oldest kubectl running this check should be 1.13

neolit123: kubectl 1.12 has it, but i don't know when it was added.

fabriziopandini:
@neolit123 all changes are addressed.
Most notably I also shifted from running the check from the node being upgraded instead of always running the check from the bootstrap control-plane node

neolit123:
thanks, let's see if this helps.
/lgtm

PR #1813, opened 2019-09-30
delete  go.mod  file in the root  directory

tanjunchen:


tanjunchen:
@neolit123

tanjunchen:
@hwdef

neolit123:
@tanjunchen please explain why is this needed right now?

tanjunchen:
@neolit123    https://github.com/kubernetes/kubernetes  
 kubernetes has go.mod and go.sum file.
kubernetes/kubeadm has  only go.mod file . Go get kubeadm, each time the local go.sum file will be generated

tanjunchen:
@neolit123 If you think it is not right. I can close the pr.  :)

neolit123:
i think it's not right, because the root kubeadm folder is not a go module. the only go module is /kinder.
we should even remove the /go.mod file because i don't think it's currently needed.

tanjunchen:
@neolit123  Thank You . I  Will Change it . I will delete the go.mod file  int the root directory If you Need

ereslibre:
/lgtm
/approve
/hold

Agree that this toplevel `go.mod` is not necessary. Holding just in case I'm missing something, @fabriziopandini feel free to unhold.

neolit123:
thanks
/holld canbcel

PR #1823, opened 2019-10-05
tests/e2e/manifests: fix correct usage of go modules

neolit123:
- fix env. var name
- call `go mod init` on each execution to get latest api-machniery
- cleanup go.{mod|sum} on exit

/assign @yagonobre @yastij 
fixes https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-all#periodic-manifest-lists

### DIFF
--- a/tests/e2e/manifests/verify_manifest_lists.sh
+++ b/tests/e2e/manifests/verify_manifest_lists.sh
@@ -11,6 +11,9 @@ echo $SEP
 
 set -x
 
+# cleanup go.* on exit
+trap "rm go.mod go.sum" EXIT
+
 # install curl if missing
 if ! `curl --version > /dev/null`; then
 	apt-get update || exit 1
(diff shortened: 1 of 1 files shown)


tests/e2e/manifests/verify_manifest_lists.sh
```diff
@@ -26,20 +29,11 @@ fi
...
-popd
-
-# set GOPATH to the local directory
-export GOMODULE=on
+# use go modules. this forces using the latest k8s.io/apimachinery package.
+export GO111MODULE=on
+go mod init
 
 # run unit tests
-go test -v ./$VERSION_PATH/version.go ./$VERSION_PATH/version_test.go
```
neolit123: these are tested in k8s CI already.

tests/e2e/manifests/verify_manifest_lists.sh
```diff
@@ -26,20 +29,11 @@ fi
...
-pushd $VERSION_PATH
-curl -sS "https://raw.githubusercontent.com/kubernetes/apimachinery/master/pkg/util/version/version_test.go" > "version_test.go"
-curl -sS "https://raw.githubusercontent.com/kubernetes/apimachinery/master/pkg/util/version/version.go" > "version.go"
-popd
-
-# set GOPATH to the local directory
-export GOMODULE=on
+# use go modules. this forces using the latest k8s.io/apimachinery package.
+export GO111MODULE=on
+go mod init
```
fabriziopandini: Ideally we should have a go.mod file and pin to a version, but being the dependency super simple I guess we are fine also in this way

neolit123: the non-pining was intentional, because this also allows us to catch changes in the apimachinery version functions.

arguably not needed, but if this starts breaking constantly we can switch to a fixed go.mod.

fabriziopandini: OK, thanks for the clarification 
Might be worth adding a comment in the code to keep track of the underlying reason of this at the first opportunity

PR #1824, opened 2019-10-07
kubeadm operator - inital-commit

fabriziopandini:
This PR creates the kubeadm operator folder, adds a kubebuilder project skeleton and a set of hacks to be user for creating a kubeadm-operator-verify prepull job

/assign @neolit123
/cc @ereslibre  
/cc @yagonobre


operator/hack/boilerplate.go.txt
```diff
@@ -0,0 +1,14 @@
+/*
+
```
ereslibre: Add regular Kubernetes authors and date

operator/hack/fetch_bins.sh
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env bash
+# Copyright 2018 The Kubernetes Authors.
```
ereslibre:
```suggestion
# Copyright 2019 The Kubernetes Authors.
```

go.mod
```diff
@@ -0,0 +1,3 @@
+module k8s.io/kubeadm
+
+go 1.12
```
ereslibre: Do we want this toplevel `go.mod`?

neolit123: we should't need it, as long as the sub-folders have a go.mod.

operator/hack/fetch_bins.sh
fabriziopandini: done

operator/hack/boilerplate.go.txt
fabriziopandini: this is the output of kubebuilder init, and it is used for autogenerated files only.

I guess that keeping out of the way regular Kubernetes authors and date simplifies management, but eventually we can fix it in follow up PRs

go.mod
fabriziopandini: We don't care about this file, but it is generated by go every time you do an operation involving go modules in kinder or in the operator folder

operator/hack/boilerplate.go.txt
ereslibre: Yup, I know what it is used for, thought having it with correct contents would be nice nevertheless, but nothing blocking.

ereslibre:
Thanks @fabriziopandini!

/lgtm

go.mod
neolit123: > We don't care about this file, but it is generated by go every time you do an operation involving go modules in kinder or in the operator folder

that's not entirely true.
if you run `cd kinder && ./hack/verify-deps.sh` the file will not be generated.

`/` itself is not a module and we should avoid treating it as such.

neolit123: what operations are causing the file generation?

fabriziopandini: I find it added every time a new dependency is added

neolit123:
looks like we merged the test-infra PR without enabling docker-in-docker.
missing parts are:
https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-cluster-lifecycle/kubeadm-kinder-discovery.yaml#L8

and

https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-cluster-lifecycle/kubeadm-kinder-discovery.yaml#L33-L34

PR #1833, opened 2019-10-14
kinder: fix flakiness due to delay on publishing artifacts

fabriziopandini:
/assign @neolit123 
/area testing
/area kinder
/priority critical-urgent

### DIFF
--- a/kinder/pkg/extract/extract.go
+++ b/kinder/pkg/extract/extract.go
@@ -34,11 +34,13 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 
 	K8sVersion "k8s.io/apimachinery/pkg/util/version"
+	"k8s.io/apimachinery/pkg/util/wait"
 	kindfs "sigs.k8s.io/kind/pkg/fs"
 )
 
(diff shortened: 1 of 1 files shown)


kinder/pkg/extract/extract.go
```diff
@@ -482,14 +484,37 @@ func saveVersionFile(addVersionFileToDst bool, dst string, version *K8sVersion.V
...
+	Duration: 1000 * time.Millisecond,
+	Factor:   2.0,
+	Jitter:   0.1,
+}
+
 func httpGet(uri string) (int64, io.ReadCloser, error) {
-	resp, err := http.Get(uri)
+	var lastError error
+	var resp *http.Response
+	err := wait.ExponentialBackoff(httpGetBackoff, func() (bool, error) {
```
neolit123: as mentioned on slack, i would have preferred the other option to use `ci/latest-bazel` temporary.
this will mask the current k8s build problems.

yet the retry is a nice to have in general and we ultimately care about tests passing.
from my experiments ~60 seconds should be enough, but let's see how this goes.

/lgtm

neolit123:
@fabriziopandini looks like it might need more than one minute:
https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-e2e-kubeadm-kinder-upgrade-1-16-master/1184257345740345346/artifacts/task-01-add-kubernetes-versions-log.txt

PR #1834, opened 2019-10-15
kubeadm-operator: first draft of the API

fabriziopandini:
This PR creates the kubeadm operator API / first draft
See https://youtu.be/0RM1ExqGl8U for the online review session

/kind feature
/sig cluster-lifecycle
/area kubeadm-operator

/priority important-soon

/cc @neolit123
/cc @ereslibre
/cc @yagonobre
/cc @yastij


operator/api/v1alpha1/runtimetask_types.go
```diff
@@ -0,0 +1,150 @@
...
+	s.CommandProgress = fmt.Sprintf("%d/%d", s.CurrentCommand, len(commands))
+}
+
+// SetCompletionTime is a utility method for setting the CompletionTime field to Now.
+func (s *RuntimeTaskStatus) SetCompletionTime() {
+	now := metav1.Now()
+	s.CompletionTime = &now
+
+	// ensure all the status attributes are clean after complete
+	s.Paused = false
```
neolit123: nit:
it feels a bit odd that the complication time setter method updates the running state (paused) of the Status object.

operator/config/rbac/role.yaml
```diff
@@ -0,0 +1,68 @@
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  creationTimestamp: null
+  name: manager-role
```
neolit123: should the file name be manager-role.yaml too?
so that it better indicates that this is about managing operator* resources?

operator/controllers/suite_test.go
```diff
@@ -0,0 +1,86 @@
...
+*/
+
+package controllers
+
+import (
+	"path/filepath"
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
```
neolit123: we shouldn't do the `.` here.
if CAP* are doing it, it's still wrong. ;)

operator/controllers/suite_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+	By("bootstrapping test environment")
+	testEnv = &envtest.Environment{
+		CRDDirectoryPaths: []string{filepath.Join("..", "config", "crd", "bases")},
+	}
+
+	var err error
+	cfg, err = testEnv.Start()
+	Expect(err).ToNot(HaveOccurred())
+	Expect(cfg).ToNot(BeNil())
```
neolit123: gomega.Expect(...

operator/controllers/suite_test.go
```diff
@@ -0,0 +1,86 @@
+/*
```
neolit123: do you think that /operator should have more unit tests in general?
we only have suite_test.go in the whole initial draft.

operator/api/v1alpha1/runtimetaskgroup_create_strategy.go
```diff
@@ -0,0 +1,41 @@
...
+
+package v1alpha1
+
+// RuntimeTaskGroupCreatePolicy is a string representation of a RuntimeTaskGroup create policy.
+type RuntimeTaskGroupCreatePolicy string
+
+const (
+	// RuntimeTaskGroupCreateSequentialStrategy forces the RuntimeTaskGroup controller to create RuntimeTasks in sequential order.
+	// New RuntimeTask are created only after the current RuntimeTime is completed successfully.
+	RuntimeTaskGroupCreateSequentialStrategy = RuntimeTaskGroupCreatePolicy("Sequential")
```
neolit123: Should be called `Serial` IMO
serial vs parallel

operator/api/v1alpha1/runtimetaskgroup_node_filter.go
```diff
@@ -0,0 +1,51 @@
...
+const (
+	// RuntimeTaskGroupNodeFilterAll forces the RuntimeTaskGroup controller to use all the nodes
+	// returned by the NodeSelector.
+	RuntimeTaskGroupNodeFilterAll = RuntimeTaskGroupNodeFilter("All")
+
+	// RuntimeTaskGroupNodeFilterHead forces the RuntimeTaskGroup controller to use only the first node
+	// returned by the NodeSelector.
+	RuntimeTaskGroupNodeFilterHead = RuntimeTaskGroupNodeFilter("First")
+
+	// RuntimeTaskGroupNodeFilterTail forces the RuntimeTaskGroup controller to use the the secondary nodes
```
neolit123: double `the`

operator/api/v1alpha1/runtimetaskgroup_node_filter.go
```diff
@@ -0,0 +1,51 @@
...
+type RuntimeTaskGroupNodeFilter string
+
+const (
+	// RuntimeTaskGroupNodeFilterAll forces the RuntimeTaskGroup controller to use all the nodes
+	// returned by the NodeSelector.
+	RuntimeTaskGroupNodeFilterAll = RuntimeTaskGroupNodeFilter("All")
+
+	// RuntimeTaskGroupNodeFilterHead forces the RuntimeTaskGroup controller to use only the first node
+	// returned by the NodeSelector.
+	RuntimeTaskGroupNodeFilterHead = RuntimeTaskGroupNodeFilter("First")
```
neolit123: `Head` and `Tail` IMO sound better than First and Second

operator/api/v1alpha1/runtimetaskgroup_node_filter.go
```diff
@@ -0,0 +1,51 @@
...
+
+package v1alpha1
+
+// RuntimeTaskGroupNodeFilter is a string representation of a RuntimeTaskGroup node filter.
+type RuntimeTaskGroupNodeFilter string
+
+const (
+	// RuntimeTaskGroupNodeFilterAll forces the RuntimeTaskGroup controller to use all the nodes
+	// returned by the NodeSelector.
+	RuntimeTaskGroupNodeFilterAll = RuntimeTaskGroupNodeFilter("All")
```
neolit123: should these strings be lowercased as the rest of the constants in the DIFF?

operator/config/crd/patches/cainjection_in_operations.yaml
```diff
@@ -0,0 +1,8 @@
+# The following patch adds a directive for certmanager to inject CA into the CRD
+# CRD conversion requires k8s 1.13 or later.
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    certmanager.k8s.io/inject-ca-from: $(CERTIFICATE_NAMESPACE)/$(CERTIFICATE_NAME)
```
neolit123: can you please explain in detail how this boiler plate works and why is it needed?

operator/config/crd/patches/webhook_in_runtimetasks.yaml
```diff
@@ -0,0 +1,17 @@
+# The following patch enables conversion webhook for CRD
+# CRD conversion requires k8s 1.13 or later.
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
```
neolit123: CRDs went GA in 1.16, any idea what the plans for kubebuilder are for moving to generating v1 manifests?

operator/config/crd/patches/webhook_in_runtimetasks.yaml
```diff
@@ -0,0 +1,17 @@
...
+  conversion:
+    strategy: Webhook
+    webhookClientConfig:
+      # this is "\n" used as a placeholder, otherwise it will be rejected by the apiserver for being blank,
+      # but we're going to set it later using the cert-manager (or potentially a patch if not using cert-manager)
+      caBundle: Cg==
+      service:
+        namespace: system
+        name: webhook-service
+        path: /convert
```
neolit123: what is our upgrade / conversion story for user API objects using conversion webhooks?
i.e. we need to support proper conversion.

AFAIK, CAPI did not support that very well between v1alpha1 -> 2

operator/go.mod
```diff
@@ -3,7 +3,11 @@ module k8s.io/kubeadm/operator
 go 1.12
 
 require (
+	github.com/go-logr/logr v0.1.0
+	github.com/onsi/ginkgo v1.6.0
+	github.com/onsi/gomega v1.4.2
 	k8s.io/apimachinery v0.0.0-20190404173353-6a84e37a896d
 	k8s.io/client-go v11.0.1-0.20190409021438-1a26190bd76a+incompatible
+	k8s.io/utils v0.0.0-20190506122338-8fab8cb257d5
```
neolit123: with kubeadm, kinder and operator we are *probably* going to end up in a weird situation where each projects needs a different version of libs like api-machinery / client-go. ;)

operator/config/rbac/role.yaml
fabriziopandini: this is autogenerated by kubebuilder; I should investigate if the file name can be changed, but I would advice against this because then we should adapt all the chain up for kustomize.

BTW. The user want see this file name because all the yaml files under config/ should be packed into a unique file during the release

neolit123: ok, understood.

operator/api/v1alpha1/runtimetask_types.go
fabriziopandini: As per comment, this is for ensuring a clean state in corner cases; might be we can get rid of this when implementing the controller

operator/controllers/suite_test.go
fabriziopandini: everything under controllers folder is autogenerated by kubebuilder and not modified by this PR
In a follow-up PR I will get rid of ginko for simpler and faster standard unit tests

operator/controllers/suite_test.go
fabriziopandini: [1]

operator/controllers/suite_test.go
fabriziopandini: definitely
see [1]

operator/api/v1alpha1/runtimetaskgroup_node_filter.go
fabriziopandini: The right casing is upper case, fixed

operator/config/crd/patches/cainjection_in_operations.yaml
fabriziopandini: Well this is part of the kubebuilder generated code, and there is a lot going on there (see https://book.kubebuilder.io/reference/reference.html). If you want I will be happy to organize another walk through sharing what I have learned so far 

Please note that some of these staff will be removed later on when we will start working on controllers, but this is out of the scope of this PR

operator/config/crd/patches/webhook_in_runtimetasks.yaml
fabriziopandini: I have to check plan with the kubebuilder folks, but being those generated files I guess we will get this for free when upgrading the controller-gen dependency/make manifest script

operator/config/crd/patches/webhook_in_runtimetasks.yaml
fabriziopandini: v1alpha1->2 was a disruptive change because e.g. the API group changed

For "normal" changes, you can leverage on conversion webhooks https://book.kubebuilder.io/multiversion-tutorial/conversion.html (same approach CAPI will use for v1alpha2 -> v1alpha3)

neolit123: > For "normal" changes, you can leverage on conversion webhooks https://book.kubebuilder.io/multiversion-tutorial/conversion.html (same approach CAPI will use for v1alpha2 -> v1alpha3)

ack

operator/go.mod
fabriziopandini: weel they are three separate projects that we are hosting in the same repo, so I don't see this so weird.

the alternative, have a unique go.mod file for kubeadm, kubeadm-operator and kinder is even less attractive to me because it mixes up things which are very different, but we should consider simplifying hacks, makefiles, godep, builds at a later stage when also kubeadm lands in this repo

neolit123: they are separate projects, true.
one potential convergence is the operator moving into kubeadm.

operator/api/v1alpha1/operation_descriptor_types.go
```diff
@@ -0,0 +1,54 @@
...
+// RenewCertificatesOperationSpec provide declarative support for the kubeadm upgrade workflow.
+type RenewCertificatesOperationSpec struct {
+
+	// INSERT ADDITIONAL SPEC FIELDS -
+	// Important: Run "make" to regenerate code after modifying this file
+}
+
+// CustomOperationSpec enable definition of custom list of RuntimeTaskGroup.
+type CustomOperationSpec struct {
+	Workflow []RuntimeTaskGroup `json:"workflow,omitempty"`
```
yastij: we should remove `omitempty`

operator/api/v1alpha1/operation_descriptor_types.go
```diff
@@ -0,0 +1,54 @@
...
+
+	// CustomOperation enable definition of custom list of RuntimeTaskGroup.
+	// +optional
+	CustomOperation *CustomOperationSpec `json:"custom,omitempty"`
+}
+
+// UpgradeOperationSpec provide declarative support for the kubeadm upgrade workflow.
+type UpgradeOperationSpec struct {
+	// +optional
+	KubernetesVersion string `json:"kubernetesVersion,omitempty"`
```
yastij: misses comment

yastij:
I can open one and add some ideas we can discuss. we can iterate based on this PR

fabriziopandini:
@neolit123 any objection to unhold?

neolit123:
@fabriziopandini looks like only me and @yastij reviewed this pretty big PR, but if you think it's good to go, please go ahead.

ereslibre:
Thank you @fabriziopandini!

I have to admit I skimmed through it, but in order to build upon it I think it's better if we get it in and start messing with it so we can have more specific discussions.

/lgtm
/hold cancel

PR #1837, opened 2019-10-17
Add err handling

yuxiaobo96:
Signed-off-by: yuxiaobo <yuxiaobogo@163.com>

### DIFF
--- a/tests/e2e/manifests/verify_manifest_lists.go
+++ b/tests/e2e/manifests/verify_manifest_lists.go
@@ -230,7 +230,9 @@ func getFromURLTimeoutSize(url string, timeout int, sizeOnly bool) (string, int,
 	if err != nil {
 		return "", -1, err
 	}
-	defer resp.Body.Close()
+	defer func() {
+		_ = resp.Body.Close()
+	}()
 
 	if resp.StatusCode != 200 {
 		return "", -1, fmt.Errorf("responded with status: %d", resp.StatusCode)


tests/e2e/manifests/verify_manifest_lists.go
```diff
@@ -230,7 +230,9 @@ func getFromURLTimeoutSize(url string, timeout int, sizeOnly bool) (string, int,
 	if err != nil {
 		return "", -1, err
 	}
-	defer resp.Body.Close()
+	defer func() {
+		_ = resp.Body.Close()
+	}()
```
neolit123: why is this change needed?
does a linter complain about the lack of return value?

yuxiaobo96: I just saw a document with similar error handling in it.
https://travix.io/unhandled-errors-in-go-3f341f2704dd
I am very sorry, maybe I didn't understand.

neolit123: no problem, but given golint is not reporting this as error i don't think we should include the change.

even if `resp.Body.Close()` fails we don't need to handle the error it will return.
so `defer resp.Body.Close()` seems fine.

yuxiaobo96: Ok, I understand it.
Thank you for your patience to explain this.
:)

PR #1840, opened 2019-10-17
kinder: use a hardcoded manifest when applying CNI 

neolit123:
- introduce pkg/data/calicomanifests.go that has CalicoCNI3_8_2.
- in pkg/cluster/manager/actions/kubeadm-init.go pass this manifest
to kubectl using STDIN.

there are two more commits:
- one that fixes kubectl apply problem from
https://github.com/kubernetes/kubeadm/pull/1838
- fix typo in calico manifest that breaks CI

### DIFF
--- a/kinder/pkg/cluster/manager/actions/kubeadm-init.go
+++ b/kinder/pkg/cluster/manager/actions/kubeadm-init.go
@@ -25,12 +25,14 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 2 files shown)


neolit123:
/assign @yagonobre @ereslibre @yastij 
looking for an LGTM because our test signal has been red for a couple of days.

neolit123:
@ereslibre 

> was docs.projectcalico.org flaky at some point?

the calico docs manifest links include the latest patch release of a MINOR, which is fine but this time a patch introduced a regression (3.8.3) and our signal broke for ~3 days. also got a complain that the SIG mailing list received too many error notifications.

> In any case, I think it's much better if we avoid relying on third parties for these kind of operations that can impact our signal.

+1

PR #1848, opened 2019-10-21
kubeadm-operator: implement the operation controller

fabriziopandini:
This PR implements the operation controller

NB. I'm splitting operator implementation in a series of PRs to make things easier to review, but the backside of this is that there are some TODO's, that will be addressed in follow up PRs when all the operators will be in places.

If this can help. we can arrange for a shared PR review

/kind feature
/sig cluster-lifecycle
/area kubeadm-operator

/priority important-soon

/cc @neolit123
/cc @ereslibre
/cc @yagonobre
/cc @yastij


operator/controllers/operation_controller.go
```diff
@@ -18,36 +18,461 @@ package controllers
...
+		name = current.Name
+	}
+	return &taskGroupMatch{
+		name:    name,
+		planned: planned,
+		current: current,
+	}
+}
+
+type taskGroupMatchList struct {
```
yastij: this should reside in a file outside of the controller

operator/controllers/operation_controller.go
```diff
@@ -18,36 +18,461 @@ package controllers
...
+		operation.Status.SetCompletionTime()
+	}
+
+	// otherwise, proceed creating TaskGroup
+
+	// if nil, set the Operation start time
+	if operation.Status.StartTime == nil {
+		operation.Status.SetStartTime()
+
+		//TODO: add a signature so we can detect if someone/something changes the operations while it is processed
```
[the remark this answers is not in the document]
fabriziopandini: I was thinking to a hash as well, but IMO it should be in the status so it can't be easily changed

fabriziopandini:
@yastij thanks for the feedback!
Everything should be addressed

operator/controllers/operation_controller.go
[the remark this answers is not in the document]
yastij: Status is better location, I agree

operator/controllers/operation_controller.go
```diff
@@ -20,34 +20,338 @@ import (
...
+			log.Error(err, "failed to patch Operation")
+			if rerr == nil {
+				rerr = err
+			}
+		}
+	}()
+
+	// Reconcile the Operation
+	if err := r.reconcileOperation(operation, log); err != nil {
+		if requeueErr, ok := errors.Cause(err).(capierrors.HasRequeueAfterError); ok {
```
ereslibre: This is including `capi` types, do we want to do this?

operator/controllers/operation_controller.go
```diff
@@ -20,34 +20,338 @@ import (
...
 
 	return ctrl.Result{}, nil
 }
 
-// SetupWithManager configures the controller for calling the reconciler
-func (r *OperationReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	return ctrl.NewControllerManagedBy(mgr).
-		For(&operatorv1alpha1.Operation{}).
-		Complete(r)
+func (r *OperationReconciler) reconcileDamonSet(operation *operatorv1.Operation, log logr.Logger) error {
```
ereslibre:
```suggestion
func (r *OperationReconciler) reconcileDaemonSet(operation *operatorv1.Operation, log logr.Logger) error {
```

and callers :)

operator/controllers/operation_controller.go
```diff
@@ -20,34 +20,338 @@ import (
...
+	if err := r.Client.Get(ctx, req.NamespacedName, operation); err != nil {
+		if apierrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
 
-	// your logic here
+	// Ignore the Operation if it is already completed or failed
+	if operation.Status.CompletionTime != nil {
```
ereslibre: If it failed, will we get through here? (would `CompletionTime` be `!= nil`?) Because `reconcileDaemonSet` would then delete the daemonset. Asking because a failed daemon set could have valuable information for debugging.

fabriziopandini: In case of an error we are not setting CompletionTime, so you have full access to the DaemonSet logs.

operator/controllers/operation_controller.go
fabriziopandini: ATM we are using also other utils from CAPI; I would defer forking/duplicating this code to a later stage.

ereslibre:
Thank you @fabriziopandini!

/lgtm
/hold

Holding to give a chance for other reviewers, feel free to unhold.

fabriziopandini:
/hold cancel
No further comments in the last 5 days, moving on

Issue #1864, opened 2019-10-31
E2E kubeadm upgrade 1-16 master failing


fabriziopandini:
This test is failing in the step task-08-cluster-info-after due to a change in etcd:

```
kinder-upgrade-control-plane:$ kubectl --kubeconfig=/etc/kubernetes/admin.conf exec -n=kube-system etcd-kinder-upgrade-control-plane -- etcdctl --endpoints=https://127.0.0.1:2379 --ca-file=/etc/kubernetes/pki/etcd/ca.crt --cert-file=/etc/kubernetes/pki/etcd/peer.crt --key-file=/etc/kubernetes/pki/etcd/peer.key member list
Error: unknown flag: --ca-file
```

PR #1865, opened 2019-10-31
fix E2E upgrade 1.16 master failures

fabriziopandini:
Fixes https://github.com/kubernetes/kubeadm/issues/1864

### DIFF
--- a/kinder/pkg/cluster/manager/actions/cluster-info.go
+++ b/kinder/pkg/cluster/manager/actions/cluster-info.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 
 	"k8s.io/kubeadm/kinder/pkg/cluster/status"
+	"k8s.io/kubeadm/kinder/pkg/constants"
 )
 
 // CluterInfo actions prints the summary information about the cluster: list of nodes,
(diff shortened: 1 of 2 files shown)


kinder/pkg/cluster/manager/actions/cluster-info.go
```diff
@@ -52,12 +53,27 @@ func CluterInfo(c *status.Cluster) error {
...
 		// local etcd is listening on localhost and on the advertise address; we are
 		// using localhost to accommodate both the use cases
-		if err := cp1.Command(
-			"kubectl", "--kubeconfig=/etc/kubernetes/admin.conf", "exec", "-n=kube-system", fmt.Sprintf("etcd-%s", c.BootstrapControlPlane().Name()),
+
+		etcdArgs := []string{
+			"--kubeconfig=/etc/kubernetes/admin.conf", "exec", "-n=kube-system", fmt.Sprintf("etcd-%s", c.BootstrapControlPlane().Name()),
 			"--",
 			"etcdctl", fmt.Sprintf("--endpoints=https://127.0.0.1:2379"),
-			"--ca-file=/etc/kubernetes/pki/etcd/ca.crt", "--cert-file=/etc/kubernetes/pki/etcd/peer.crt", "--key-file=/etc/kubernetes/pki/etcd/peer.key",
```
neolit123: this is quite the breaking change in `etcdctl`, but we should stop executing etcdctl from the etcd container, in general.
etcdctl will soon be removed from from the etcd image (due to distroless).

so the member list has to be obtained in another way.

neolit123: it's unclear, but it's already WIP.
might happen in 2020 or later.

Issue #1867, opened 2019-10-31
operator: organize imports


fabriziopandini:
As per comment [here](https://github.com/kubernetes/kubeadm/pull/1861#discussion_r341181565) we would like to change import organization from

two groups
- import from the standard library
- everything else

to three groups
- everything else
- k8s.io sigs.k8s.io  imports
- kubeadm/operator imports.

this should be done on all the files under /operator folder

hypnoglow:
To which one of three groups should third-party imports (like `github.com/pkg/errors`) go? I guess the second group?

fabriziopandini:
/close
via https://github.com/kubernetes/kubeadm/pull/1869

PR #1869, opened 2019-10-31
operator: organize imports

hypnoglow:
This changeset makes operator imports sorted as described in #1867.

I also added `-local k8s.io/kubeadm/operator` to **goimports** command, so it will enforce making the 3rd group of imports belonging to "kubeadm/operator". I don't know if this flag is appreciated across kubernetes org; if it's not, I can remove that.

### DIFF
--- a/operator/controllers/operation_controller.go
+++ b/operator/controllers/operation_controller.go
@@ -24,13 +24,14 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/client-go/tools/record"
-	operatorv1 "k8s.io/kubeadm/operator/api/v1alpha1"
-	operatorerrors "k8s.io/kubeadm/operator/errors"
-	"k8s.io/kubeadm/operator/operations"
 	capierrors "sigs.k8s.io/cluster-api/errors"
 	"sigs.k8s.io/cluster-api/util/patch"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	operatorv1 "k8s.io/kubeadm/operator/api/v1alpha1"
+	operatorerrors "k8s.io/kubeadm/operator/errors"
+	"k8s.io/kubeadm/operator/operations"
 )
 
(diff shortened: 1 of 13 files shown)


operator/controllers/operation_controller.go
```diff
@@ -24,13 +24,14 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
```
fabriziopandini: I think import from github.com should go in the first group (they are not k8s.io/sigs-k8s.io)
@yastij opinions ^^^ ?

operator/controllers/suite_test.go
```diff
@@ -25,11 +25,12 @@ import (
 
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
-	operatorv1 "k8s.io/kubeadm/operator/api/v1alpha1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
```
fabriziopandini: move "github.com/onsi/gomega" and  "github.com/onsi/gomega" in the first group?

operator/controllers/operation_controller.go
hypnoglow: @fabriziopandini  btw goimports does not allow grouping std imports with third-party ones, so I guess these imports have to go to the 2nd group

neolit123: BTW we should be following what kubernetes/kubernetes is doing for go formatting.
the tool for that is `go fmt` and not `goimports`:
https://github.com/kubernetes/kubernetes/blob/master/hack/update-gofmt.sh

hypnoglow: @neolit123 If so, what's the decision here? Do we then even need to use 3 groups of imports, if there is no agreement in k/k on this, and if we can't force this via automation?

neolit123: this repo already has CI with gofmt enabled, so it will catch problems. but i think it's still possible to use multiple groups.

fabriziopandini:
This is passing ci checks, so I make this merge to keep going with the POC
We can eventually reiterate in follow up PRs on 
- go fmt vs go imports
- operator/controllers/operation_controller.go
- operator/controllers/suite_test.go

/lgtm

PR #1873, opened 2019-11-01
kinder: remove "sigs.k8s.io/kind/pkg/log" dependency

fabriziopandini:
Little cleanup trying to make easier future upgrades of the kind release


neolit123:
LGTM, but looks like my comment got lost when i sent it here and also this now has conflicts.

fabriziopandini:
@neolit123 rebased. Let me know if there is something else to address

PR #1876, opened 2019-11-01
kinder: remove kind config

fabriziopandini:
This is relevant simplification in kinder (112 additions and 737 deletions)

kinder currently supports the users passing a v1alpha3 kind config to `kinder create`, but
- only few settings are actually supported by actions 
- most of the config management moved to internal in kind, so it was necessary to fork it some time ago and now most of the code lives into kinder

This PR removes the `--config` from `kinder create`; This does not have impact on the kubeadm E2E test, and should have only a limited impact on users because all the supported features are already mirrored by other flags.

After removing the flag, it is possible to clean up the forked code and make some simplifications that are acceptable for the kinder target use cases (kubeadm developers/kubeadm E2E).
1. kinder does not support custom address/port number for the API server exposed on the host (it always use 127.0.0.1 and a random port)
2. kinder does not support custom host mounts and custom port mappings

Please note that 2. enables further simplifications in follow-up PR which should address the impact of "sigs.k8s.io/kind/pkg/cluster/nodes" moved to internal in the master branch of kind  😞


neolit123:
this is an interesting dependency decoupling. we are indeed not using the config and maintaining it is some level of burden. are some of the recent changes leading us towards using only the logic for the kind base and node image from kind and kinder becoming this separate tool for node creation and topology actions? CAPD ended up only forking a small subset of kind.

/lgtm
/hold
feel free to unhold

fabriziopandini:
>are some of the recent changes leading us towards using only the logic for the kind base and node image from kind and kinder becoming this separate tool for node creation and topology actions? CAPD ended up only forking a small subset of kind.

@neolit123 I'm checking kind at master, and the majority of the library command we are using are being moved to internal. This and the fact we want to preserve docker support are the two major forces pushing us in this direction.

BTW CAPD is not a great example, because it relies on CAPI/CAPBK for bootstrap, while for kinder we need to implement our own actions which are specifically designed for kubeadm developers/kubeadm E2E

fabriziopandini:
as per comment above
/hold cancel

PR #1887, opened 2019-11-05
kinder: add a test job family for running regular tests

neolit123:
Currently these are done using kind and the kind kubetest deployer.
We are switching these test jobs to kinder for consistency
and better control.

The workflows do the following:
- pull a base image
- add artifacts for the target version
- create a 3CP2W cluster
- run cluster-info, e2e-kubeadm and e2e
- export logs, reset, delete the cluster

xref https://github.com/kubernetes/kubernetes/issues/84676
test-infra PR: https://github.com/kubernetes/test-infra/pull/15130

### DIFF
--- a/kinder/ci/workflows/regular-tasks.yaml
+++ b/kinder/ci/workflows/regular-tasks.yaml
@@ -1,0 +1,135 @@
+# IMPORTANT! this workflow is imported by regular-* workflows.
+version: 1
+summary: |
+  This workflow implements a sequence of tasks used test the proper functioning
+  of kubeadm version X with Kubernetes version X.
+vars:
+  # vars defines default values for variable used by tasks in this workflow;
+  # those values might be overridden when importing this files.
+  kubernetesVersion: v1.13.5
+  controlPlaneNodes: 3
+  workerNodes: 2
+  baseImage: kindest/base:v20191105-ee880e9b # has containerd
+  image: kindest/node:test
+  clusterName: kinder-regular
+tasks:
+- name: pull-base-image
+  description: |
+    pulls kindest/base image with docker in docker and all the prerequisites necessary for running kind(er)
+  cmd: docker
+  args:
+    - pull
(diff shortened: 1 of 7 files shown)


kinder/ci/workflows/regular-tasks.yaml
```diff
@@ -0,0 +1,135 @@
...
+  of kubeadm version X with Kubernetes version X.
+vars:
+  # vars defines default values for variable used by tasks in this workflow;
+  # those values might be overridden when importing this files.
+  kubernetesVersion: v1.13.5
+  controlPlaneNodes: 3
+  workerNodes: 2
+  baseImage: kindest/base:v20190403-1ebf15f
+  image: kindest/node:test
+  clusterName: kinder-xony
```
fabriziopandini: kinder-regular?

kinder/ci/workflows/regular-tasks.yaml
```diff
@@ -0,0 +1,135 @@
...
+  cmd: kinder
+  args:
+    - export
+    - logs
+    - --loglevel=debug
+    - --name={{ .vars.clusterName }}
+    - "{{ .env.ARTIFACTS }}"
+  force: true
+  timeout: 5m
+  # kind export log is know to be flaky, so we are temporary ignoring errors in order
```
fabriziopandini: This should not be the case anymore, but I'm fine in keeping this (potential source of flakes)

neolit123: ok, let's remove these all at once as we have this comment in multiple YAML files.

kinder/ci/workflows/regular-tasks.yaml
neolit123: updated, also added https://github.com/kubernetes-sigs/kind/blob/master/pkg/build/node/node.go#L46
as the base image (from latest kind master) let's see if it works.

PR #1892, opened 2019-11-07
kinder: use the etcd tag version for defining etcdctl flags

neolit123:
etcdctl needs different cert arguments for 3.4+ and older.
Execute EtcdImage() on the bootstrap control plane to obtain the
active etcd image version. This should solve issues where after
upgrade or skew tests the version of etcdctl does not match
the /kind/version version.

The alternative is to add means to update /kind/version for
certain workflows and "do" actions.


neolit123:
go 1.13 flake?
/retest

PR #1896, opened 2019-11-08
kinder: experiment with writing the full stderr to junit Failures

neolit123:
Looking at existing junit files:
https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-e2e-kubeadm-kinder-kustomize-master/1192600800178737153/artifacts/junit_runner.xml

`<failure>` tags always(?) contain "exit status 1" and this
is not very helpful.

To provide a better indication of the failure also write
the stderr contents to the failure.

This requires the following changes:
- use CDATA for the failure tag
- use MultiWriter with a new bytes.Buffer()
- pass the buffer contents as "failure" tag
- in case of failure pass the error as a new argument to
registerTestCase() and return this error instead of the failure string

this change is experimental.

example output:
```
<?xml version="1.0" encoding="UTF-8"?>
<testsuite failures="1" tests="1" time="3.014406431">
    <testcase classname="kinder.test.workflow" name="task-00-bash" time="3.013926316">
        <failure><![CDATA[some stderr text

exit status 1]]></failure>
    </testcase>
```

### DIFF
--- a/kinder/pkg/test/workflow/taskCmdRunner.go
+++ b/kinder/pkg/test/workflow/taskCmdRunner.go
@@ -17,6 +17,7 @@ limitations under the License.
 package workflow
 
 import (
+	"bytes"
 	"encoding/xml"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


neolit123:
/hold
i think this will also include klog output which isn't great.
need to test it more.

kinder/pkg/test/workflow/taskCmdRunner.go
```diff
@@ -155,8 +172,8 @@ func (c *taskCmdRunner) Run(t *taskCmd, artifacts string, verbose bool) error {
 		cleanup(t.Cmd)
 
 		// otherwise record test case failure and exits with error
-		return c.registerTestCase(t.Name,
-			withFailure(err.Error()),
+		return c.registerTestCase(t.Name, err,
```
fabriziopandini: might be we can also check if some sort of nested error?

neolit123: you mean checking if an error is wrapped using the errors.Wrap calls?

PR #1897, opened 2019-11-09
kinder: fix more etcd image related issues

neolit123:
xref https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-kubeadm#kubeadm-kinder-upgrade-1-16-1-17

```
kinder: do not cache the kubernetesVersion  …
The call is cheap and caching can present issues when the version
is changed in the container.
 
kinder: call etcdctl flags based on the etcdctl version  …
- call etcdctl from the etcd container and get its version
- append the version specific etcdctl cert flags based on the version
```

### DIFF
--- a/kinder/pkg/cluster/status/node.go
+++ b/kinder/pkg/cluster/status/node.go
@@ -47,6 +47,7 @@ type Node struct {
 	ipv6            string
 	cri             ContainerRuntime
 	kubeadmVersion  *K8sVersion.Version
+	etcdImage       string
 	skip            bool
 	commandMutators []commandMutator
 }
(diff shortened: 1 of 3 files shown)


kinder/pkg/cluster/status/node.go
```diff
@@ -40,17 +40,16 @@ type commandMutator = func(*cmd.ProxyCmd) *cmd.ProxyCmd
...
-	etcdImage         string
-	skip              bool
-	commandMutators   []commandMutator
+	name            string
+	role            string
+	ports           map[int32]int32
+	ipv4            string
+	ipv6            string
+	cri             ContainerRuntime
+	kubeadmVersion  *K8sVersion.Version
```
fabriziopandini: I'm going to open an issue to remove also this from the cache, so everything can change on a node is not cached anymoore

PR #1903, opened 2019-11-10
kinder: get the etcdctl version from the etcd binary

neolit123:
"etcdctl version" is not portable to 3.3.x.
Looks like the download instructions are wrong too:
https://github.com/etcd-io/etcd/releases/tag/v3.3.17

etcdctl 3.3.x uses "--version" and 3.4.x uses "version".
To not a have a chicken and egg problem get the version
from the etcd binary which is consistent across 3.3.x and 3.4.x -
uses "--version".


kinder/pkg/cluster/manager/actions/cluster-info.go
```diff
@@ -78,13 +77,14 @@ func CluterInfo(c *status.Cluster) error {
 			return err
 		}
 
+		fmt.Printf("Using etcdctl version: %s\n", etcdctlVersion)
```
fabriziopandini: `fmt.Printf` --> `cp1.Infof`

PR #1904, opened 2019-11-10
kinder: add a presubmit CI workflow

neolit123:
It performs an upgrade from ci/latest to ci/latest.

The e2e and e2e-kubeadm tasks are in dry-run mode, because
the goal of this test is to check for kinder errors and not
kubeadm / k8s errors.

test infra PR: https://github.com/kubernetes/test-infra/pull/15217

### DIFF
--- a/kinder/ci/workflows/presubmit-upgrade-master.yaml
+++ b/kinder/ci/workflows/presubmit-upgrade-master.yaml
@@ -1,0 +1,135 @@
+version: 1
+summary: |
+  This workflow tests upgrades to Kubernetes ci/latest and also a set of other actions for pull requests
+  name      > pull-kubeadm-kinder-upgrade-master
+  config    > https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/kubeadm/kubeadm-presubmits.yaml
+vars:
+  initVersion: "{{ resolve `ci/latest` }}"
+  upgradeVersion: "{{ resolve `ci/latest` }}"
+  controlPlaneNodes: 3
+  workerNodes: 2
+  baseImage: kindest/base:v20190403-1ebf15f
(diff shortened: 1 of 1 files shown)


kinder/ci/workflows/presubmit-upgrade-beta-master.yaml
```diff
@@ -0,0 +1,135 @@
+version: 1
+summary: |
+  this workflow tests kubeadm upgrades from Kubernetes version ci/k8s-beta (the latest unstable release) to version ci/latest
+  test grid > https://testgrid.k8s.io/sig-cluster-lifecycle-kubeadm#pull-kubeadm-kinder-upgrade-beta-latest
+  config    > https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/kubeadm/kubeadm-presubmits.yaml
+vars:
+  initVersion: "{{ resolve `ci/k8s-beta` }}"
+  upgradeVersion: "{{ resolve `ci/latest` }}"
```
fabriziopandini: I'm not fully aware of the differences between `latest` label series and the `k8s` label series, but I find a little bit weird to mix the two. Then, considering we are using `k8s-beta`, I guess this should be `k8s-master`

neolit123: k8s-beta can give a different version from latest
k8s-master is always the same as latest.

i will do `ci/latest` for both, but the upgrade will be a NO-OP.

neolit123:
> However, as soon as the release is cut, we should consider if/how to split periodic jobs and pre-submit jobs into two separated folders because the number of files under workflow is starting to get relevant

SGTM

neolit123:
updated.

PR #1910, opened 2019-11-12
kinder: improve workflow verification

neolit123:
- make verify-workflow.go use logrus, to be consistent with the
underlying API functions it's calling.
- dry-run a workflow as part of verification
- allow passing a io.Writer to workflow.Run, when calling from
kinder use os.Stdout, when called from the verify-workflow pass
ioutil.Discard
- print better message in verify-workflows.sh
- handle missing template vars via "missingkey=error"
- use logrus.Debug for shadowed variables

### DIFF
--- a/kinder/pkg/test/workflow/workflow.go
+++ b/kinder/pkg/test/workflow/workflow.go
@@ -28,6 +28,7 @@ package workflow
 
 import (
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -216,7 +217,7 @@ func (w *Workflow) expandImports(file string) error {
 				w.Vars[k] = v
 				continue
 			}
-			log.Infof("var %s in workflow file %s is shadowed by var %[1]s in parent workflow file %[3]s", k, path, file)
+			log.Debugf("var %s in workflow file %s is shadowed by var %[1]s in parent workflow file %[3]s", k, path, file)
 		}
 
 		// merge the env vars from the import file into the parent file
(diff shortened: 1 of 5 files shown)


kinder/pkg/test/workflow/workflow.go
```diff
@@ -216,7 +217,7 @@ func (w *Workflow) expandImports(file string) error {
 				w.Vars[k] = v
 				continue
 			}
-			log.Infof("var %s in workflow file %s is shadowed by var %[1]s in parent workflow file %[3]s", k, path, file)
+			log.Warningf("var %s in workflow file %s is shadowed by var %[1]s in parent workflow file %[3]s", k, path, file)
```
fabriziopandini: [1] I think we should downgrade this as a debug message because we are heavily relying on variable shadowing, and this message make s the output of verify-workflows really verbose without adding much value. wdyt?

kinder/pkg/test/workflow/workflow.go
```diff
@@ -226,7 +227,7 @@ func (w *Workflow) expandImports(file string) error {
 				w.Env[k] = v
 				continue
 			}
-			log.Infof("env var %s in workflow file %s is shadowed by env var %[1]s in parent workflow file %[3]s", k, path, file)
+			log.Warningf("env var %s in workflow file %s is shadowed by env var %[1]s in parent workflow file %[3]s", k, path, file)
```
fabriziopandini: [1]

kinder/pkg/test/workflow/workflow.go
neolit123: ok, will revert to infof

neolit123:
@fabriziopandini updated.

PR #1919, opened 2019-11-15
kubeadm operator: add unit test tests - part1

fabriziopandini:
This PR adds the first set of unit test for the kubeadm operator

### DIFF
--- a/operator/controllers/runtimetask_controller_test.go
+++ b/operator/controllers/runtimetask_controller_test.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package controllers
 
 import (
@@ -6,8 +22,9 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/record"
-	"k8s.io/kubeadm/operator/errors"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/log"
+
+	"k8s.io/kubeadm/operator/errors"
 
 	operatorv1 "k8s.io/kubeadm/operator/api/v1alpha1"
 )
--- a/operator/controllers/runtimetask_controller_test.go
+++ b/operator/controllers/runtimetask_controller_test.go
@@ -1,0 +1,803 @@
+package controllers
+
+import (
+	"reflect"
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/kubeadm/operator/errors"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/log"
+
+	operatorv1 "k8s.io/kubeadm/operator/api/v1alpha1"
+)
+
+func TestRuntimeTaskReconciler_reconcilePhase(t *testing.T) {
+	tx := timePtr(metav1.Now())
+
+	type args struct {
+		task *operatorv1.RuntimeTask
+	}
+	type want struct {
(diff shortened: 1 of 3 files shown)


operator/controllers/runtimetask_controller_test.go
```diff
@@ -0,0 +1,820 @@
...
+	type args struct {
+		task *operatorv1.RuntimeTask
+	}
+	type want struct {
+		task *operatorv1.RuntimeTask
+	}
+	tests := []struct {
+		name string
+		args args
+		want want
```
neolit123: as mentioned on the @SataQiu PR this should be, like so:
https://github.com/kubernetes/kubeadm/pull/1923/files#diff-1175c7d3cbb2ee3af06c7f1bbd008d60R34-R36

operator/controllers/runtimetask_controller_test.go
```diff
@@ -0,0 +1,820 @@
...
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			r := &RuntimeTaskReconciler{}
+			r.reconcilePhase(tt.args.task)
+
+			if !reflect.DeepEqual(tt.args.task, tt.want.task) {
+				t.Errorf("reconcilePhase() = %v, want %v", tt.args.task, tt.want.task)
```
neolit123: we should follow this pattern:
https://github.com/kubernetes/kubeadm/pull/1923/files#diff-1175c7d3cbb2ee3af06c7f1bbd008d60R134

including the function name is not needed.

operator/controllers/runtimetask_reconcile_test.go
```diff
@@ -0,0 +1,264 @@
...
+			},
+		},
+	}
+
+	got := reconcileTasks(nodes, tasks)
+
+	// All
+	want := []string{"node1-a", "node1-b", "node1-c", "node1-d", "node2", "node3", "node4"}
+	if got := taskNames(got.all); !reflect.DeepEqual(got, want) {
+		t.Errorf("newTaskGroupChildProxy().all = %v, want %v", got, want)
```
neolit123: this list of "all", "pending"...etc, might be simplified with reflection and a table, but will be more difficult to read.

operator/controllers/runtimetask_reconcile_test.go
```diff
@@ -0,0 +1,264 @@
...
+	"reflect"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	operatorv1 "k8s.io/kubeadm/operator/api/v1alpha1"
+)
+
+func Test_newTaskGroupChildProxy(t *testing.T) {
```
neolit123: i think we shouldn't use underscores in function names

operator/controllers/runtimetask_controller_test.go
fabriziopandini: [1] I don't see any value added in renaming args in input and want in output, so I don't think this should block the PR

TBH I have to say that the PR from @SataQiu deviated from the example provided, without discussing the point with the author of the issue

operator/controllers/runtimetask_controller_test.go
fabriziopandini: I don't see any problem in having the name of the function in the error message; actually I think that having the function name can be useful in detecting and fixing the problem

operator/controllers/runtimetask_reconcile_test.go
fabriziopandini: "all", "pending" etc are all part of the same struct, and all of them are set by the same call. Having a table will make it really difficult to read

operator/controllers/runtimetask_reconcile_test.go
fabriziopandini: Same as [1] I don't see any value added in renaming, so I don't think this should block the PR

neolit123:
leaving review and LGTM to @SataQiu 
/cc @SataQiu

fabriziopandini:
@SataQiu, kind ping

SataQiu:
I‘d like to merge this PR first. From a functional point of view, this PR is fine.
If you don't mind, I will send another PR after this merged to make all the test code have the same style.

/lgtm

SataQiu:
@fabriziopandini 
The CI failed! This is because the function name conflicts (It's my fault).
I have sent a PR to resolve this issue #1942

fabriziopandini:
@SataQiu 

> If you don't mind, I will send another PR after this merged to make all the test code have the same style.

Feel free to send such PR. My point is only to not block at this stage of this effort and to iterate quickly. Everything that goes in this direction is more than welcome!

PR #1937, opened 2019-11-24
Operator: add more tests, fix some errors, small refactors and cleanups

fabriziopandini:
This is a follow up of https://github.com/kubernetes/kubeadm/pull/1919, with more tests, fix some errors detected while testing, small refactors and cleanups


operator/controllers/operation_controller.go
```diff
@@ -245,64 +243,71 @@ func (r *OperationReconciler) reconcileNormal(operation *operatorv1.Operation, l
 	operation.Status.FailedGroups = int32(len(taskGroups.failed))
 	operation.Status.InvalidGroups = int32(len(taskGroups.invalid))
 
+	return taskGroups, nil
+}
+
+func (r *OperationReconciler) reconcileNormal(operation *operatorv1.Operation, taskGroups *taskGroupReconcileList, log logr.Logger) error {
+	// If the Operation doesn't have finalizer, add it.
```
yastij: is this code needed ? If not let's just deleted it, we can always add it back later

operator/controllers/runtimetaskgroup_controller.go
```diff
@@ -192,47 +185,58 @@ func (r *RuntimeTaskGroupReconciler) reconcileNormal(executionMode operatorv1.Op
 	taskgroup.Status.FailedNodes = int32(len(tasks.failed))
 	taskgroup.Status.InvalidNodes = int32(len(tasks.invalid))
 
+	return tasks, nil
+}
+
+func (r *RuntimeTaskGroupReconciler) reconcileNormal(executionMode operatorv1.OperationExecutionMode, taskgroup *operatorv1.RuntimeTaskGroup, tasks *taskReconcileList, log logr.Logger) error {
+	// If the TaskGroup doesn't have finalizer, add it.
+	//if !util.Contains(taskgroup.Finalizers, operatorv1alpha1.TaskGroupFinalizer) {
```
yastij: same

operator/controllers/util.go
```diff
@@ -233,6 +233,9 @@ func createDaemonSet(c client.Client, operation *operatorv1.Operation, namespace
 
 	} else {
 		// Expose /metrics on default (insecure) port
+		if daemonSet.Annotations == nil {
+			daemonSet.Annotations = map[string]string{}
```
yastij: nit: can't we just do this anyway when initializing the ds first ?

operator/controllers/operation_controller.go
fabriziopandini: [1] I'm still defining if we need finalizers or not.
I prefer to keep the comment around so I'm reminded about it

operator/controllers/runtimetaskgroup_controller.go
fabriziopandini: [1]

operator/controllers/util.go
fabriziopandini: Annotations are used only in this branch of code, and I would prefer to keep the branch auto-consistent.
But no problem to change if this is blocking for you

fabriziopandini:
@yastij @SataQiu kindly ping

fabriziopandini:
/honk
(I like this command!)

PR #1945, opened 2019-11-30
Kinder: remove-kind-exec

fabriziopandini:
Cleanup trying to make easier future upgrades of the kind release

Now Host cmds and Node cmds share are both implemented in `knder/pkg/exec` and share the same semantics (ndr Before only  Node cmds was implemented in kinder, while host cmd was still relying on kind)

/assign @neolit123 
/area kinder
/kind cleanup
/priority backlog
/milestone next


fabriziopandini:
/hold
until release cut

PR #1952, opened 2019-12-02
kinder: add --volume option to configure mount volumes on containers

fabriziopandini:
After removing support for the kind config, kinder lost some features.
Some users showed interest in mount volumes (e.g. for mounting sources folders inside containers) and this PR is an MVP for restoring mount volumes.

/assign @neolit123
/area kinder
/kind cleanup
/priority backlog
/milestone next

/hold
until release cut

### DIFF
--- a/kinder/cmd/kinder/create/cluster/createcluster.go
+++ b/kinder/cmd/kinder/create/cluster/createcluster.go
@@ -25,23 +25,19 @@ import (
 )
 
 const (
-	configFlagName               = "config"
-	controlPlaneNodesFlagName    = "control-plane-nodes"
-	workerNodesFLagName          = "worker-nodes"
-	kubeDNSFLagName              = "kube-dns"
-	externalEtcdFlagName         = "external-etcd"
-	externalLoadBalancerFlagName = "external-load-balancer"
+	controlPlaneNodesFlagName = "control-plane-nodes"
+	workerNodesFlagName       = "worker-nodes"
 )
 
 type flagpole struct {
 	Name                 string
-	Config               string
 	ImageName            string
 	Workers              int
 	ControlPlanes        int
 	Retain               bool
 	ExternalEtcd         bool
 	ExternalLoadBalancer bool
+	Volumes              []string
 }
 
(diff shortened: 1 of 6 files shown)


kinder/cmd/kinder/create/cluster/createcluster.go
```diff
@@ -25,23 +25,19 @@ import (
...
 
 const (
-	configFlagName               = "config"
-	controlPlaneNodesFlagName    = "control-plane-nodes"
-	workerNodesFLagName          = "worker-nodes"
-	kubeDNSFLagName              = "kube-dns"
-	externalEtcdFlagName         = "external-etcd"
-	externalLoadBalancerFlagName = "external-load-balancer"
+	controlPlaneNodesFlagName = "control-plane-nodes"
+	workerNodesFLagName       = "worker-nodes"
```
neolit123: FLag -> Flag

kinder/cmd/kinder/create/cluster/createcluster.go
```diff
@@ -83,14 +79,19 @@ func NewCommand() *cobra.Command {
...
 	)
 	cmd.Flags().BoolVar(
 		&flags.ExternalLoadBalancer,
-		externalLoadBalancerFlagName, false,
+		"external-load-balancer", false,
 		"add an external load balancer to the cluster (implicit if number of control-plane nodes>1)",
 	)
+	cmd.Flags().StringSliceVarP(
+		&flags.Volumes,
+		"volume", "v", nil,
```
neolit123: i would keep `volume` only for this instead of `v` too.
kind moved to using `--v` for klog-like verbosity levels.
we might do that in the future too.

kinder/cmd/kinder/create/cluster/createcluster.go
fabriziopandini: fixed 😄

kinder/cmd/kinder/create/cluster/createcluster.go
fabriziopandini: +1

PR #1957, opened 2019-12-06
kinder: use a short list of ignored preflight errors

neolit123:
scope down the list of ignored preflight errors.

### DIFF
--- a/kinder/pkg/constants/constants.go
+++ b/kinder/pkg/constants/constants.go
@@ -73,6 +73,10 @@ const (
 	// TODO: send a PR to define this value in a kind constant (currently it is not)
 	KubeadmConfigPath = "/kind/kubeadm.conf"
 
+	// KubeadmIgnorePreflightErrorsFlag holds the default list of preflight errors to skip
+	// on "kubeadm init" and "kubeadm join"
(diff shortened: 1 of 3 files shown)


kinder/pkg/constants/constants.go
```diff
@@ -73,6 +73,10 @@ const (
 	// TODO: send a PR to define this value in a kind constant (currently it is not)
 	KubeadmConfigPath = "/kind/kubeadm.conf"
 
+	// KubeadmIgnorePreflightErrorsFlag holds the default list of preflight errors to skip
+	// on "kubeadm init" and "kubeadm join"
+	KubeadmIgnorePreflightErrorsFlag = "--ignore-preflight-errors=Swap,SystemVerification,ImagePull,FileContent--proc-sys-net-bridge-bridge-nf-call-iptables"
```
neolit123: with `ci/latest` the `ImagePull` check fails.
this PR already merged:
https://github.com/kubernetes/kubernetes/pull/85965/files

so i'm going to assume that we were ignoring these errors before too.

ideally we should not ignore this preflight check.
possibly, eventually making kubeadm support `ImagePullPolicy` as per the discussion with ncdc.

/hold

neolit123:
@fabriziopandini i don't see the image pull warnings in the latest kinder master logs, so maybe it's possible to remove this. i will try again when i get some time.

neolit123:
@fabriziopandini updated without the ImagePull skip.

Issue #1981, opened 2019-12-19
upgrade from 1.17 to master is broken


neolit123:
our e2e caught this:
https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-kubeadm#kubeadm-kinder-upgrade-1-17-master

https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-e2e-kubeadm-kinder-upgrade-1-17-master/1207767136668225537/build-log.txt

```
[upgrade/etcd] Upgrading to TLS for etcd
error execution phase control-plane: couldn't complete the static pod upgrade: etcd cluster is not healthy: failed to dial endpoint https://172.17.0.4:2379 with maintenance client: context deadline exceeded
To see the stack trace of this error execute with --v=5 or higher
Error: failed to exec action kubeadm-upgrade: exit status 1
 exit status 1
```

the k/k diff does not reveal anything suspicious:
https://github.com/kubernetes/kubernetes/compare/0b830f3d6...e5f0648c6

we don't seem to have upgrade related changes in kinder during that period but this needs to be double checked. EDIT: the runs before and after the first failures use the same k/kubeadm commit 8622735.

related k/k tickets:
https://github.com/kubernetes/kubernetes/pull/86689
https://github.com/kubernetes/kubernetes/issues/86676
https://github.com/kubernetes/kubernetes/issues/86873

/kind failing-test
/priority important-soon

neolit123:
here are some logs with --v=6

```
kinder-upgrade-control-plane-2:$ kubeadm upgrade node --v=6
time="01:11:18" level=debug msg="Running: [docker exec kinder-upgrade-control-plane-2 kubeadm upgrade node --v=6]"
I1220 01:11:18.390660    7220 loader.go:375] Config loaded from file:  /etc/kubernetes/kubelet.conf
[upgrade] Reading configuration from the cluster...
[upgrade] FYI: You can look at this config file with 'kubectl -n kube-system get cm kubeadm-config -oyaml'
I1220 01:11:18.412954    7220 round_trippers.go:443] GET https://172.17.0.7:6443/api/v1/namespaces/kube-system/configmaps/kubeadm-config?timeout=10s 200 OK in 9 milliseconds
I1220 01:11:18.418317    7220 round_trippers.go:443] GET https://172.17.0.7:6443/api/v1/namespaces/kube-system/configmaps/kube-proxy?timeout=10s 200 OK in 3 milliseconds
[...]

PR #1989, opened 2019-12-27
kinder: upgrade node's kubelet and kubectl last

neolit123:
- Upgrade node kubelet and kubectl last, after the kubeadm binary
is upgraded and after the "kubeadm upgrade" process has finished
on all nodes.

see
https://github.com/kubernetes/kubernetes/issues/86676
https://github.com/kubernetes/kubeadm/issues/1981

deciding on how to proceed here, after https://github.com/kubernetes/kubernetes/issues/86676 sees more discussion.

### DIFF
--- a/kinder/pkg/cluster/manager/actions/kubeadm-upgrade.go
+++ b/kinder/pkg/cluster/manager/actions/kubeadm-upgrade.go
@@ -41,8 +41,9 @@ func KubeadmUpgrade(c *status.Cluster, upgradeVersion *K8sVersion.Version, kusto
 	}
 
 	preloadUpgradeImages(c, upgradeVersion)
+	nodeList := c.K8sNodes().EligibleForActions()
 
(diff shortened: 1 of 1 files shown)


neolit123:
> Delete the kubelet checkpoint file before kubelet restart
because in 1.18 the kubelet started panicing (again?) because
of it.

this part is a bug in the kubelet and we must not do it from kinder.
i will update the PR once the related kubelet PRs merge.

fabriziopandini:
> i will update the PR once the related kubelet PRs merge

got it. ping me after the PR is updated for approval

neolit123:
@fabriziopandini 
i have updated this PR

after https://github.com/kubernetes/kubernetes/issues/86676 closed, we no longer have to manually clean the checkpoint file. the PR only changes the upgrade order for the kubelet & kubectl.

/hold cancel

Issue #1918, opened 2019-11-14
kinder: investigate flakes related to /kinder/cluster-settings.yaml


neolit123:
kinder writes the file /kinder/cluster-settings.yaml on nodes and it contains cluster settings.
we are seeing flakes related to that with no useful error output:

https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-kubeadm-kinder-master/1195019721611677696

```
time="16:48:03" level=error msg="failed to write cluster settings to node kinder-regular-control-plane-1: failed to write /kinder/cluster-settings.yaml: exit status 1"
Error: failed to create cluster: failed to write cluster settings to node kinder-regular-control-plane-1: failed to write /kinder/cluster-settings.yaml: exit status 1
```

/kind flake
/assign
/priority important-longterm

neolit123:
@fabriziopandini 

FYI, looks like https://github.com/kubernetes/kubeadm/pull/1946 did not fix this issue.

https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-e2e-kubeadm-kinder-master/1205249784051404810/build-log.txt

> Error: failed to create cluster: failed to write cluster settings to node kinder-regular-control-plane-3: failed to write /kinder/cluster-settings.yaml: exit status 1

fabriziopandini:
@neolit123 is there an issue for tracking this problem?

neolit123:
this is the issue.

fabriziopandini:
:rofl: got lost in translation!
As soon as possible I will try to give it a try and make it surface a better error message

neolit123:
no worries.

> As soon as possible I will try to give it a try and make it surface a better error message

was also planning to do that myself and try to reproduce it locally.

neolit123:
https://github.com/kubernetes/kubeadm/commit/5e91aed7406f2e9a62e3034a32834c03ecf4ca5e#diff-38f115aa3b3fc261d4510c63088d75a3
did not help much in terms of revealing why the `docker cp` is failing from time to time.
https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-e2e-kubeadm-kinder-master/1211187265594724353/artifacts/task-02-create-cluster-log.txt

fabriziopandini:
@neolit123 I have sent a temporary fixe. Feel free to consider or ignore it

neolit123:
this is also failing for `node-settings`:
[...]

PR #1996, opened 2020-01-03
Kinder: Temporarily disable clustersettings

fabriziopandini:
This is a temporary fix for https://github.com/kubernetes/kubeadm/issues/1918

/assign @neolit123

### DIFF
--- a/kinder/pkg/cluster/status/node.go
+++ b/kinder/pkg/cluster/status/node.go
@@ -231,46 +231,57 @@ func (n *Node) EtcdImage() (string, error) {
 	return n.etcdImage, nil
 }
 
-const clusterSettingsPath = "/kinder/cluster-settings.yaml"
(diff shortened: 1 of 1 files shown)


neolit123:
@fabriziopandini i didn't even look what cluster-settings is doing.
are they mandatory? can we use a different method for sharing settings?

the verify job is failing on the PR.

fabriziopandini:
@neolit123 

> are they mandatory?
After the last round of changes the scope of cluster setting is now very limited, so no problem in disabling them.

> can we use a different method for sharing settings?
Writing a file in a container is used in many places in kinder, so it should not be a problem for cluster setting as well. It only requires some time to investigate the problem but this is not a priority ...

neolit123:
/lgtm

> It only requires some time to investigate the problem but this is not a priority ...

i've spent quite some time on this already and couldn't find the cause. it's not easy to reproduce.

PR #2073, opened 2020-03-24
Null point check

Hellcatlk:


kinder/third_party/kind/loadbalancer/config.go
```diff
@@ -64,6 +64,9 @@ backend kube-apiservers
 // Config returns a kubeadm config generated from config data, in particular
 // the kubernetes version
 func Config(data *ConfigData) (config string, err error) {
+	if 0 > data.ControlPlanePort {
```
neolit123: hi, there is no way for kinder to expose the port configuration.
why do you think this check is needed?

also it should be written as `data.ControlPlanePort <= 0`

operator/commands/factory.go
```diff
@@ -27,51 +27,52 @@ import (
 
 // RunCommand run the command on a node
 func RunCommand(c *operatorv1.CommandDescriptor, log logr.Logger) error {
-
```
neolit123: please send changes for kinder and the operator in separate PRs.

operator/commands/factory.go
```diff
@@ -27,51 +27,52 @@ import (
...
-	if c.Fail != nil {
-		time.Sleep(5 * time.Second)
-		return errors.New("command fail failed")
-	}
-
-	if c.Wait != nil {
-		time.Sleep(time.Duration(c.Wait.Seconds) * time.Second)
-		return nil
+	if c != nil {
```
neolit123: this should be:
```
if c == nil {
    return early...
}
```

but why do you think `c` can end up as `nil` here?

operator/operations/factory.go
```diff
@@ -24,34 +24,37 @@ import (
...
-		return setupRenewCertificates(), nil
-	}
-
-	if operation.Spec.Upgrade != nil {
-		return setupUpgrade(), nil
-	}
-
-	if operation.Spec.CustomOperation != nil {
-		return setupCustom(), nil
+	if operation != nil {
```
neolit123: same as [1]

operator/operations/factory.go
```diff
@@ -24,34 +24,37 @@ import (
...
 	}
 
 	return nil, errors.New("Invalid Operation.Spec.OperatorDescriptor. There are no operation implementation matching this spec")
 }
 
 // TaskGroupList return the list of TaskGroup to be performed by an operation
 func TaskGroupList(operation *operatorv1.Operation) (*operatorv1.RuntimeTaskGroupList, error) {
-	if operation.Spec.RenewCertificates != nil {
-		return planRenewCertificates(operation, operation.Spec.RenewCertificates), nil
+	if operation != nil {
```
neolit123: [1]

PR #2089, opened 2020-03-29
[kinder] remove dependency kind kustomize

fabriziopandini:
This PR remove dependency from "sigs.k8s.io/kind/pkg/kustomize" by forking some code from older versions of kind.

Please note that kind moved away from kustomize in favor or lighter solutions. In kinder, we are not taking this step yet because the size of the binary is not a major concern at this stage.

/area kinder
/assign @neolit123

### DIFF
--- a/kinder/pkg/kubeadm/kustomize.go
+++ b/kinder/pkg/kubeadm/kustomize.go
@@ -1,0 +1,122 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


kinder/pkg/kubeadm/kustomize.go
```diff
@@ -0,0 +1,122 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
neolit123: 2020

fabriziopandini: ooops...

PR #2090, opened 2020-03-29
[kinder] remove dependency yaml.v2

fabriziopandini:
this PR remove dependency from `gopkg.in/yaml.v2` since we are already using `sigs.k8s.io/yaml` in other places

/area clusterctl
/assign @neolit123


fabriziopandini:
/hold 
it seems there are differences in the serialization of `time.Duration` 🤔

neolit123:
> it seems there are differences in the serialization of time.Duration

seems like so.

neolit123:
sigs.k8s.io/yaml uses encoding/json as it's JSON backend. so this will work.
https://play.golang.org/p/GbC4QLJCPCG
it needs some special handling.

neolit123:
@fabriziopandini i will send the PR for the above.
/close

PR #2099, opened 2020-04-05
remove dependency kind concurrent

fabriziopandini:
this PR removes another kind dependency
 
/area kinder
/assign @neolit123

### DIFF
--- a/kinder/pkg/cluster/manager/create.go
+++ b/kinder/pkg/cluster/manager/create.go
@@ -27,7 +27,6 @@ import (
 	"k8s.io/kubeadm/kinder/pkg/constants"
 	"k8s.io/kubeadm/kinder/pkg/cri"
 	"k8s.io/kubeadm/kinder/pkg/exec"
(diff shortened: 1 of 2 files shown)


kinder/pkg/cluster/manager/create.go
```diff
@@ -296,3 +295,23 @@ func ensureNodeImage(image string) {
 	// we don't care if this errors, we'll still try to run which also pulls
 	_, _ = kinddocker.PullIfNotPresent(image, 4)
 }
+
+// UntilError runs all funcs in separate goroutines, returning the
+// first non-nil error returned from funcs, or nil if all funcs return nil
+// Nb. this func was originally imported from "sigs.k8s.io/kind/pkg/concurrent"; it is still available
+// in the kind codebase, but it has been slightly refactored.
+func untilError(funcs []func() error) error {
```
neolit123, line 303: we might have to move this in a utils package when we need it in more places.
/lgtm

PR #2101, opened 2020-04-05
cleanup third_party kubeadm

fabriziopandini:
The code under third_party was forked from kind when some library moved to internal.
Now time is passed, the gap between kind and kinder is widening, so it is time to start to maintain this code as part of the kinder codebase.

/area kinder
/assign @neolit123


fabriziopandini:
Rebased...

PR #2108, opened 2020-04-20
kinder: add e2e workflow for testing join and upgrade without addon ConfigMap

neolit123:
Add a workflow that:
 - creates 1CP1W cluster
 - inits the CP node
 - deletes the kube-proxy and coredns config maps
 - joins the worker
 - upgrades the CP node (to the same version as init)

This ensures that kubeadm tolerates missing addon config maps.
"upgrade apply" still requires skipping the CoreDNS preflight checks,
but this is controllable by the user.

Skipping the "addon" phase from "init" completely results in a cluster
that does not work, as a proxy / CNI solution has to be deployed for the
"upgrade-prepull*" DS to complete. Potentially replacement addons
can provide a better e2e test for this (in the future).


neolit123:
@fabriziopandini i've added the new entry in `kinder/ci/kubeadm-periodic.tests.md`

PR #2118, opened 2020-04-27
fix kinder smoke test

### LINKED ISSUE #2117: `kinder do smoke-test` fails
## What keywords did you search in kubeadm issues before filing this one?
smoke

## Is this a BUG REPORT or FEATURE REQUEST?
BUG REPORT

## Versions

**kubeadm version** (use `kubeadm version`):
`kubeadm version: &version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-03-25T20:54:24Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}`

## What happened?
I was following the kinder [getting started](https://github.com/kubernetes/kubeadm/blob/master/kinder/doc/getting-started.md) guide.  I was able to create a cluster, but the `kinder do smoke-test` fails.

## What you expected to happen?
Expected the smoke test to pass.

## How to reproduce it (as minimally and precisely as possible)?
Follow along the kinder [getting started](https://github.com/kubernetes/kubeadm/blob/master/kinder/doc/getting-started.md) guide.

## Anything else we need to know?
The smoke test runs -

- `kubectl --kubeconfig=/etc/kubernetes/admin.conf run nginx ...` which creates an nginx pod.
- `kubectl --kubeconfig=/etc/kubernetes/admin.conf expose deployment nginx --port=80 --type=NodePort` which fails with "deployments.apps "nginx" not found"

prasadkatti:
- Replace `kubectl run` with `kubectl create deployment`
- Wait for service to be reachable before trying to access it

fixes #2117

### DIFF
--- a/kinder/pkg/cluster/manager/actions/smoke-test.go
+++ b/kinder/pkg/cluster/manager/actions/smoke-test.go
@@ -41,7 +41,7 @@ func SmokeTest(c *status.Cluster, wait time.Duration) error {
 	if err := cp1.Command(
 		"kubectl",
 		"--kubeconfig=/etc/kubernetes/admin.conf",
-		"run", "nginx", "--image=nginx:1.15.9-alpine", "--image-pull-policy=IfNotPresent",
+		"create", "deployment", "nginx", "--image=nginx:1.15.9-alpine",
 	).RunWithEcho(); err != nil {
 		return err
 	}
@@ -51,7 +51,7 @@ func SmokeTest(c *status.Cluster, wait time.Duration) error {
 	}
 
 	// Test service type NodePort
-	cp1.Infof("service type NodePort")
+	cp1.Infof("test NodePort service")
 
 	if err := cp1.Command(
 		"kubectl",
@@ -66,9 +66,11 @@ func SmokeTest(c *status.Cluster, wait time.Duration) error {
 		return err
 	}
 
-	err = checkNodePort(c, nodePort)
-	if err != nil {
-		return err
+	for _, n := range c.K8sNodes() {
+		err = waitForNodePort(c, n, 30*time.Second, nodePort)
+		if err != nil {
+			return err
+		}
 	}
 
 	podName, err := getPodName(cp1, "nginx")
@@ -141,44 +143,11 @@ func getNodePort(n *status.Node, svc string) (string, error) {
 	return strings.Trim(lines[0], "'"), nil
 }
 
-func checkNodePort(c *status.Cluster, port string) error {
-	for _, n := range c.K8sNodes() {
-		fmt.Printf("checking node port %s on node %s...", port, n.Name())
-
-		//TODO: test IPV6
-		ip, _, err := n.IP()
-		if err != nil {
-			return err
-		}
-
-		lines, err := n.Command(
-			"curl", "-Is", fmt.Sprintf("http://%s:%s", ip, port),
-		).Silent().RunAndCapture()
-
-		if err != nil {
-			return errors.Wrapf(err, "error checking node port")
-		}
-
-		if len(lines) < 1 {
-			return errors.Wrapf(err, "error checking node port. invalid answer")
-		}
-
-		if strings.Trim(lines[0], "\n\r") == "HTTP/1.1 200 OK" {
-			fmt.Printf("pass!\n")
-			continue
-		}
-
-		return errors.Errorf("node port %s on node %s doesn't works", port, n.Name())
-	}
-
-	return nil
-}
-
 func getPodName(n *status.Node, label string) (string, error) {
 	lines, err := n.Command(
 		"kubectl",
 		"--kubeconfig=/etc/kubernetes/admin.conf",
-		"get", "pods", "-l", fmt.Sprintf("run=%s", label), "-o", "jsonpath='{.items[0].metadata.name}'",
+		"get", "pods", "-l", fmt.Sprintf("app=%s", label), "-o", "jsonpath='{.items[0].metadata.name}'",
 	).Silent().RunAndCapture()
 	if err != nil {
 		return "", errors.Wrapf(err, "failed to get pod name")
--- a/kinder/pkg/cluster/manager/actions/waiter.go
+++ b/kinder/pkg/cluster/manager/actions/waiter.go
@@ -54,6 +54,18 @@ func waitForPodsRunning(c *status.Cluster, n *status.Node, wait time.Duration, l
 	return nil
 }
 
+// waitForNodePort waits for a nodePort to become ready
+func waitForNodePort(c *status.Cluster, n *status.Node, wait time.Duration, nodePort string) error {
+	n.Infof("waiting for NodePort %q to become ready (timeout %s)", nodePort, wait)
+	if pass := waitFor(c, n, wait,
+		nodePortIsReady(n, nodePort),
+	); !pass {
+		return errors.New("timeout: NodePort not ready")
+	}
+	fmt.Println()
+	return nil
+}
+
 // waitNewWorkerNodeReady waits for a new control plane node reaching the target state after join
 func waitNewWorkerNodeReady(c *status.Cluster, n *status.Node, wait time.Duration) error {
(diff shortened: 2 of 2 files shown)


kinder/pkg/cluster/manager/actions/smoke-test.go
```diff
@@ -66,6 +66,9 @@ func SmokeTest(c *status.Cluster, wait time.Duration) error {
 		return err
 	}
 
+	// wait a few seconds for svc to be reachable
+	time.Sleep(3 * time.Second)
```
fabriziopandini, line 70: Is it possible to replace this with a call to kubectl wait? ( kubectl wait --for=condition=ready service/nginx --timeout=30s)

neolit123: ^ +1

prasadkatti: Thanks for the feedback. Looks like `kubectl wait` can't wait for service to be ready yet (https://github.com/kubernetes/kubernetes/issues/80828). There is an open issue for this - https://github.com/kubernetes/kubernetes/issues/83094.

fabriziopandini: ookkk...
Considering the smoke test command is not used in CI, I think we can go with this for now,  but please add a TODO in the code to get this fixed in the future
@neolit123 opinions?

neolit123: TODO for now seems fine.

the proper way to solve this would be to add a new function here:
https://github.com/kubernetes/kubeadm/blob/83de8de3dbb3a3bb40e17d6df3638bec4bf6f28a/kinder/pkg/cluster/manager/actions/waiter.go

```
func nodePortIsReady(c *status.Cluster, port string) func(c *status.Cluster, n *status.Node) bool {
...
}
```

and also a new `waitForNodePort(...)` (<- `checkNodePort` logic goes there) that uses `waitFor()`. see `waitForPodsRunning` for example.

prasadkatti: I can try to do this the proper way later today.

prasadkatti: I have added a waiter function. PTAL when you can. :)

kinder/pkg/cluster/manager/actions/waiter.go
```diff
@@ -255,6 +267,33 @@ func podsAreRunning(n *status.Node, label string, replicas int) func(c *status.C
 	}
 }
 
+// nodePortIsReady implements a function that tests if a nodePort is ready
+func nodePortIsReady(n *status.Node, port string) func(c *status.Cluster, n *status.Node) bool {
+	return func(c *status.Cluster, n *status.Node) bool {
+		fmt.Printf("checking node port %s on node %s...\n", port, n.Name())
```
neolit123, line 273: i think we should remove this call.
and only leave the call on line 289. it prints only on a positive outcome (return true)

kinder/pkg/cluster/manager/actions/waiter.go
```diff
@@ -54,6 +54,18 @@ func waitForPodsRunning(c *status.Cluster, n *status.Node, wait time.Duration, l
 	return nil
 }
 
+// waitForNodePort waits for a nodePort to become ready
+func waitForNodePort(c *status.Cluster, n *status.Node, wait time.Duration, nodePort string) error {
+	n.Infof("waiting for NodePort to become ready (timeout %s)", wait)
```
neolit123, line 59: potentially include nodePort in the message:

```
n.Infof("waiting for NodePort %q to become ready (timeout %s)", nodePort, wait)
```

kinder/pkg/cluster/manager/actions/smoke-test.go
```diff
@@ -66,10 +66,7 @@ func SmokeTest(c *status.Cluster, wait time.Duration) error {
 		return err
 	}
 
-	// wait a few seconds for svc to be reachable
-	time.Sleep(3 * time.Second)
-
-	err = checkNodePort(c, nodePort)
+	err = waitForNodePort(c, cp1, 30*time.Second, nodePort)
```
neolit123, line 69: the old code iterated over all the nodes.

so this should be:
```
for _, n := range c.K8sNodes() {
	err = waitForNodePort(c, n, wait, nodePort)
	if err != nil {
		return err
	}
}
```

this makes `wait` is a bit odd in the context of `SmokeTest()` as a whole, because it is now used in a couple of "wait" locations, so it can wait for a maximum of wait * 2

neolit123: one wait of handling that is passing `wait / 2` to all wait* calls.
it's default value is 5 minutes:
https://github.com/kubernetes/kubeadm/blob/cf387fefe4fd92ae6587b6a60c59d8b621a72095/kinder/cmd/kinder/do/do.go#L104

kinder/pkg/cluster/manager/actions/waiter.go
prasadkatti: okay 👍

kinder/pkg/cluster/manager/actions/waiter.go
prasadkatti: 👍

kinder/pkg/cluster/manager/actions/smoke-test.go
prasadkatti: I just kept the timeout to 30s for each of those calls. From what I saw the NodePort svc is ready within a couple of seconds. So the shorter timeout will help to fail fast.
The cluster that you get with when following the getting started guide has 3 nodes.

prasadkatti:
I had to change the label used to search for the pod from `run=nginx` to `app=nginx`. This was a result of changing `kubectl run` to `kubectl create deployment`.

kinder/pkg/cluster/manager/actions/smoke-test.go
prasadkatti: How difficult would it be to pass multiple `try` funcs to the `waitFor` so that this happens in parallel?

fabriziopandini:
/lgtm
/approve

@prasadkatti please squash commits, after that I will lift the hold
/hold

kinder/pkg/cluster/manager/actions/smoke-test.go
neolit123: waitFor already supports this, but is a bit difficult as this is the sequence ATM:
- create deployment
- wait for pod
- expose port
- wait for port

if we swap "expose port" / "wait for pod" the pod might not be ready yet and expose will fail.
so i'd say leave it the way it is.

prasadkatti:
@fabriziopandini - I have squashed the commits and force-pushed.

neolit123:
the previous failure in `pull-kubeadm-kinder-upgrade-master`  seemed like a flake in pulling artifacts, so this seems fine.

/lgtm
/approve
/hold cancel

prasadkatti:
@neolit123 @fabriziopandini - Do we want to add smoke-test to ci?

neolit123:
i think its not a bad idea.
should be added above this line https://github.com/kubernetes/kubeadm/blob/master/kinder/ci/workflows/presubmit-upgrade-master.yaml#L117

fabriziopandini:
Smoke tests are more for the Devs than for Ci; in CI we are running the compliance suite and this is far more reliable than smoke tests, so IMO it is not necessary.

neolit123:
the only benefit would be to catch regressions in the smoke test, but i see your point.

fabriziopandini:
What about having smoke test running in tests that are not already doing e2e (so use it as a light-weight alternative to e2e)

neolit123:
a separate CI job seems better.

fabriziopandini:
I was thinking at existing jobs not running e2e (e.g Kustomize if I remember well)

PR #2130, opened 2020-05-02
Update testing section in contributing guide

prasadkatti:


CONTRIBUTING.md
```diff
@@ -67,7 +67,7 @@ You can also run unit tests for specific `kubeadm` packages using:
 ```
 make test WHAT=<package> GOFLAGS="-v"
 ```
-Where `<package>` can be `./cmd/kubeadm/app/cmd`, `./cmd/kubeadm/app/utils`, `./cmd/kubeadm/app/features`, etc.
+Where `<package>` can be `./cmd/kubeadm/app/cmd`, `./cmd/kubeadm/app/util`, `./cmd/kubeadm/app/features`, etc.
```
prasadkatti, line 70: the `./hack/make-rules/test-kubeadm-cmd.sh` a couple of lines above also needs to be fixed. There is no script called `test-kubeadm-cmd.sh` in `./hack/make-rules`

neolit123: should be refactored to:

> To run `kubeadm` integration tests call:
> ```
> make test-cmd WHAT=kubeadm
> ```

prasadkatti: @neolit123 - I looked at the Makefile and the `test-cmd` target is for building and testing cmdline tests [1]. So should it be _"To run kubeadm cmdline tests"_ as opposed to _"To run kubeadm integration tests"_?

1. https://github.com/kubernetes/kubernetes/blob/b943e49e7698d307ae08057c0e3ec88ceb25c81a/build/root/Makefile#L301

neolit123: i think we should call them the "integration tests" as that's how the maintainers know them.

neolit123: or to be more specific: "the command line integration tests".

prasadkatti: Updated the section a bit. I have changed the order so that we have -

- cmd to run unit tests
- cmd to run cmdline integration tests

The section is titled "Running unit tests". So let me know if the second command needs to be in a different section of its own.

neolit123: i think it's fine for now. the whole section will potentially change once kubeadm moves outside of k/k.
/ok-to-test

neolit123:
thanks
/lgtm
/approve

PR #2131, opened 2020-05-05
kinder: adjust timeouts to attempt to fix flakes

neolit123:
xref https://github.com/kubernetes/kubernetes/issues/90761

5 minutes has become more and more flaky for some tasks.

This could be caused by a couple of factors:

1) A scheduling / node regression in core of sort. I think I started
seeing this in my local kubeadm clusters (without kind/kinder)
around 1.16-1.17.

2) Slower infra: GCE VMs / Prow Pods / CR could be involved. Maybe
our jobs have less resources compared to before or maybe it only
happens sometimes.

Increase the timeout for the following tasks to 10m:
- "kubeadm-join": The nodes just don't get ready in time for 3CP2W workflows.
- "e2e-kubeadm": It seems building the suite times-out a lot now.
Maybe the suite just got a little bigger and we were borderline
flaky without catching it before.
- "kubeadm-upgrade": it just timeouts waiting on the nodes to become
Ready. Happens for 2CP1W workflows too. This case in particular
tells me that we maybe have a case of 2) as well, somehow, occasionally.

### DIFF
--- a/kinder/ci/workflows/discovery-tasks.yaml
+++ b/kinder/ci/workflows/discovery-tasks.yaml
@@ -118,6 +118,7 @@ tasks:
     - --test-flags=--report-dir={{ .env.ARTIFACTS }}
     - --name={{ .vars.clusterName }}
     - --loglevel=debug
+  timeout: 10m
 - name: get-logs
   description: |
     Collects all the test logs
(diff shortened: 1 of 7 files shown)


neolit123:
@hasheddan ideally the PR should have waited for @fabriziopandini 's review.

if needed i can send a new amending PR.

hasheddan:
oops sorry about that @neolit123! Will keep that in mind next time!

fabriziopandini:
@neolit123 thanks for this fix. I'm observing a similar increase of timeout errors in CAPI, so I agree there should be some generalized problem ongoing on nodes/test-infra

PR #2133, opened 2020-05-07
kinder prepull images during build

neolit123:
xref https://github.com/kubernetes/kubernetes/issues/90761

first commit is just a typo fix.
second commit:
```
The idea of this change is to reduce the amount of network traffic
kinder e2e test workflows have to do with the attempt to fix flakes.

For 3CP setup etcd has to be pulled 3x~200MB (6x~200MB for upgrade).
After this change it has to be pulled 1x~200MB (2x~200MB for upgrade).

- pre-pull non-kube images for docker and containerd based on the output
of the kubeadm binary for init/join and binary for upgrade.
- pre-load these images for docker during cluster create and upgrade
- for containerd the images are not pre-loaded due to issues related
to the containerd "snapshoter" functionality. add TODOs to resolve
that in the future. one potential fix is to import them on runtime
the same way it is done for docker.
NOTE: currently we don't have e2e tests for containerd.

CNI images (Calico) are not pre-pulled.
```

/area kinder
/kind feature

### DIFF
--- a/kinder/pkg/build/alter/alter.go
+++ b/kinder/pkg/build/alter/alter.go
@@ -256,13 +256,29 @@ func fixRepository(repository string) string {
 }
 
 func (c *Context) alterImage(bitsInstallers []bits.Installer, bc *bits.BuildContext) error {
-	// create alter container
+	// get the container runtime from the base image
+	runtime, err := status.InspectCRIinImage(c.baseImage)
+	if err != nil {
+		return errors.Wrap(err, "error detecting CRI!")
+	}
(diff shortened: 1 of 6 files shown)


neolit123:
/assign @fabriziopandini 
PTAL as CI has become quite flaky and last week i saw timeouts pulling the etcd image.

/hold 
for review

fabriziopandini:
/hold cancel
/approve
/lgtm
@neolit123 thanks for this PR. I like the direction the code is taking.

PR #2134, opened 2020-05-11
fix checking of pre-loaded images

prasadkatti:
`kubeadm config images list --kubernetes-version=%s` emits a warning -
> WARNING: kubeadm cannot validate component configs for API groups [kubelet.config.k8s.io kubeproxy.config.k8s.io]

The `expected` list of images contains this warning and that results in -
>Some of the required images are not pre-loaded into the container runtime

This PR aims to fix that by redirecting the stderr of kubeadm to `/dev/null`

### DIFF
--- a/kinder/pkg/cluster/manager/actions/images.go
+++ b/kinder/pkg/cluster/manager/actions/images.go
@@ -31,9 +31,11 @@ import (
 func checkImagesForVersion(n *status.Node, version string) error {
 	n.Infof("Checking pre-loaded images")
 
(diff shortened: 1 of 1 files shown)


neolit123:
> This PR aims to fix that by redirecting the stderr of kubeadm to /dev/null

something to note is that etcd, coredns and pause are not pre-loaded.
this is the reason i've sent the following PR and will try to finish it soon:
https://github.com/kubernetes/kubeadm/pull/2133

Issue #2142, opened 2020-05-14
1.19: kinder: fix failing e2e tests


neolit123:
this change broke e2e tests:
https://github.com/kubernetes/kubeadm/pull/2133

https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-kubeadm#kubeadm-kinder-master
https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-e2e-kubeadm-kinder-master/1260480951998222336/artifacts/task-01-add-kubernetes-versions-log.txt

it says that it detected containerd in the base image.
so this might be a case where we were running docker in the PR pre-submit job that verifies PRs for kinder and it did not catch the problem.

/priority critical-urgent
/kind bug
/area kinder
/assign

neolit123:
yes so the jobs are using different base images:
https://github.com/kubernetes/kubeadm/blob/master/kinder/ci/workflows/regular-tasks.yaml#L12
https://github.com/kubernetes/kubeadm/blob/master/kinder/ci/workflows/presubmit-upgrade-master.yaml#L11

i assume `kindest/base:v20190403-1ebf15f` has docker too and that's why we didn't catch the problem.

fabriziopandini:
So, if I got this right, regular test is failing, and this is the only test using an image based on containerd
I think that is important for us to get this working with containerd, because the adoption is growing (e.g. Cluster API)
However from the logs I can't spot the problem, because the runtime detection works well but at a certain point it fails to create a folder... Any clue on what is going on?

neolit123:
i could not reproduce the problem locally, but will try again because i think i was testing with the image that also has docker. in that image docker is detected as the CR, but i overwrote that to containerd (as it also has containerd) to be able to test my change.

something to note is that e2e test on k/k (via kind) and k/cluster-api-* are all containerd, so kuibeadm e2e and maybe some sig-node tests are the only tests for dockershim.

neolit123:
hi, @hasheddan and @kubernetes/ci-signal https://github.com/kubernetes/kubeadm/pull/2143 is a potential fix for an actual bug in our e2e test tools that was introduced recently.

but me and @fabriziopandini had a discussion today and we are seeing a lot more flakes for certain tasks in our jobs. something that normally takes 2 minutes, would sometimes take 10minutes+, which implies network+CPU restrictions in the Prow cluster.

[...]

PR #2143, opened 2020-05-14
kinder fix issues around containerd base images

neolit123:
fixes https://github.com/kubernetes/kubeadm/issues/2142

tested with two separate images:
```
kinder build node-image-variant --base-image kindest/base:v20190403-1ebf15f --with-init-artifacts=./alter/bits/init --with-upgrade-artifacts=./alter/bits/upgrade --loglevel=debug
kinder build node-image-variant --base-image kindest/base:v20191105-ee880e9b --with-init-artifacts=./alter/bits/init --with-upgrade-artifacts=./alter/bits/upgrade --loglevel=debug
```

then for each image:
```
kinder create cluster ...
kinder do kubeadm-init ...
kinder do kubeadm-upgrade ...
```

commits:

```
    kinder: don't delete upgrade image tars for containerd
    
    Keep the images in the container for Init and Upgrade for both CRs.
-------
    kinder: load init images from tars on cluster creation for containerd
    
    - When pulling images with containerd export them to tars
    - Add loadImages() that uses "ctr images import"
-------
    kinder: don't use /usr/local/bin/entrypoint for alter
    
    - /usr/local/bin/entrypoint throws an error since /kind/product_name
    is missing
    - /sbin/init throws a strage docker related error
    
    Using this for alter is not needed. Use sleep infinity instead.
-------
    kinder: remove PreloadInitImages for alter
    
    Preloading images on containerd using snapshotting / ctr image pull
    does not work. Remove the function PreloadInitImages for
    both docker (was empty) and containerd.
    
    The preloading of init images would now happen for both CR on
    cluster creation.
```

### DIFF
--- a/kinder/pkg/build/alter/alter.go
+++ b/kinder/pkg/build/alter/alter.go
@@ -305,15 +305,14 @@ func (c *Context) alterImage(bitsInstallers []bits.Installer, bc *bits.BuildCont
 		}
 	}
 
-	log.Info("Pre pull extra images ...")
+	log.Info("Pre-pull extra images ...")
 	if err := alterHelper.PrePullAdditionalImages(bc, "/kind", "/kinder/upgrade"); err != nil {
 		return errors.Wrapf(err, "image build Failed! Failed to pre-pull additional images into %s", runtime)
 	}
 
-	log.Info("Pre loading images ...")
-	// TODO: preload upgrade images for containerd too?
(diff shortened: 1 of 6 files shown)


neolit123:
we still might want to add a presubmit for testing containerd, or extend the existing "upgrade" presubmit to just create two clusters one after another.

PR #2168, opened 2020-06-07
[Kinder] preload images for containerd on build

fabriziopandini:
This PR restores preload images for containerd on build

/assing @neolit123


kinder/pkg/cri/containerd/alterhelper.go
```diff
@@ -86,6 +92,15 @@ func PullImages(bc *bits.BuildContext, images []string, targetPath string) error
 	return nil
 }
 
+// PreLoadInitImages preload images required by kubeadm-init into the containerd runtime that exists inside a kind(er) node
+func PreLoadInitImages(bc *bits.BuildContext, srcFolder string) error {
+	// NB. this code is an extract from "sigs.k8s.io/kind/pkg/build/node"
+	return bc.RunInContainer(
+		"bash", "-c",
+		`find `+srcFolder+` -name *.tar -print0 | xargs -0 -n 1 -P $(nproc) ctr --namespace=k8s.io images import --all-platforms --no-unpack`,
```
neolit123, line 100: my understanding is that this preload does not work and i think you verified this during our zoom?
did you make it work somehow in this PR?

from my tests, after the container is started images are not there and they end up being pulled from the internet, which caused kubeadm CI to pull some extra gigabytes from the web.

kinder/pkg/cri/containerd/alterhelper.go
```diff
@@ -36,9 +37,6 @@ func GetAlterContainerArgs() ([]string, []string) {
 	runArgs := []string{
 		// privileged is required for "ctr image pull" permissions
 		"--privileged",
-		// the snapshot storage must be a volume.
-		// see the info in Commit()
-		"-v=/var/lib/containerd",
```
fabriziopandini, line 41: Removing this line  is the core of this change, because with "-v=/var/lib/containerd" all the pre-loaded images were lost.
everything else is simply restoring the workflow

neolit123: ack, thanks for the info.

fabriziopandini:
/hold cancel
@neolit123 PTAL
in my local test it is not pulling images for containerd; I have only a small concern for images retrieved by PrePullAdditionalImages, but this part was not changed so IMO this one is good to go

PR #2173, opened 2020-06-08
add --with-path option to kinder build node-image-variant

### LINKED ISSUE #2169: [kinder] add --with-file option to kinder build  node-image-variant
We would like to add new `--with-file sourcepath:destPath` option to `kinder build  node-image-variant`  so the user can add a new file to the image.

In order to do so:
- a new flag should be added [here](https://github.com/kubernetes/kubeadm/blob/master/kinder/cmd/kinder/build/nodevariant/nodevariant.go)
- a new option should be added [here](https://github.com/kubernetes/kubeadm/blob/master/kinder/pkg/build/alter/alter.go)
- if the new option is set, a new `fileBits` should be added to the list of `bitsInstallers` [here](https://github.com/kubernetes/kubeadm/blob/62fd18ac6c35de633a1146ce53ac99c6b01dbd31/kinder/pkg/build/alter/alter.go#L153)
- the new file bits should be implemented [here](https://github.com/kubernetes/kubeadm/tree/master/kinder/pkg/build/bits)
- the `Prepare` method should copy source path on c.HostBitsPath() /files (using a tmp file name)
- the `Install` method should copy the tmp file to destPath (inside the image)

prasadkatti:
Fixes #2169

### DIFF
--- a/kinder/cmd/kinder/build/nodevariant/nodevariant.go
+++ b/kinder/cmd/kinder/build/nodevariant/nodevariant.go
@@ -34,6 +34,7 @@ type flagpole struct {
 	Kubeadm                 string
 	Kubelet                 string
 	PrePullAdditionalImages bool
+	Path                    []string
 }
 
 // NewCommand returns a new cobra.Command for building the node image
@@ -94,6 +95,11 @@ func NewCommand() *cobra.Command {
 		true,
 		"pre-pull kubeadm additional required images such as etcd, coredns and pause, etc",
 	)
+	cmd.Flags().StringSliceVar(
+		&flags.Path, "with-path",
+		nil,
+		"sourcePath:destPath pairs; copies file/dir at sourcePath on the host to destPath inside the image, destPath has to be absolute",
+	)
 	return cmd
 }
 
@@ -111,6 +117,7 @@ func runE(flags *flagpole, cmd *cobra.Command, args []string) error {
 		alter.WithPrePullAdditionalImages(flags.PrePullAdditionalImages),
 		// bits options
 		alter.WithImageNamePrefix(flags.ImageNamePrefix),
+		alter.WithPath(flags.Path),
 	)
 	if err != nil {
 		return errors.Wrap(err, "error creating alter context")
--- a/kinder/pkg/build/bits/pathBits.go
+++ b/kinder/pkg/build/bits/pathBits.go
@@ -1,0 +1,123 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package bits
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+
+	"k8s.io/kubeadm/kinder/pkg/extract"
+)
+
+// pathBits defines a bit installer that allows to copy a file or dir (recursively) from host machine into the node image
+type pathBits struct {
+	paths []string
+}
+
+var _ Installer = &pathBits{}
+
+// NewPathBits returns a new path Installer
+func NewPathBits(paths []string) Installer {
+	return &pathBits{
+		paths: paths,
+	}
+}
+
+// Prepare implements bits.Prepare
+func (b *pathBits) Prepare(c *BuildContext) (map[string]string, error) {
+
+	// ensure the staging dest path exists on host at HostBitsPath
+	dstDir := filepath.Join(c.HostBitsPath(), "files")
+	if err := os.Mkdir(dstDir, 0777); err != nil {
+		return nil, errors.Wrap(err, "failed to make bits dir")
+	}
+
+	for idx, path := range b.paths {
+
+		pathSrcDst := strings.Split(path, ":")
+
+		if len(pathSrcDst) != 2 {
+			return nil, errors.New(fmt.Sprintf("invalid value %q for --with-path option, has to be in 'srcPath:destPath' format", path))
+		}
+
+		if !filepath.IsAbs(pathSrcDst[1]) {
+			return nil, errors.New(fmt.Sprintf("invalid value %q for --with-path option, destPath has to be an absolute path", path))
+		}
+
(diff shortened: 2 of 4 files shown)


SataQiu:
Hi @prasadkatti 
If I understand correctly, this PR should fix issue #2169, rather than pr #2172. 😆

kinder/cmd/kinder/build/nodevariant/nodevariant.go
```diff
@@ -88,6 +89,11 @@ func NewCommand() *cobra.Command {
 		"",
 		"override the kubeadm binary existing in the image with the given version/build-label/file or folder containing the kubelet binary",
 	)
+	cmd.Flags().StringSliceVar(
+		&flags.File, "with-file",
+		nil,
+		"sourcePath:destPath pairs; copies file at sourcePath on the host to destPath inside the image",
```
neolit123, line 95: should we support generic copy of paths instead of files only and potentially rename this flag to `with-path`?

/cc @fabriziopandini

kinder/pkg/build/bits/fileBits.go
```diff
@@ -0,0 +1,105 @@
...
+		return nil, errors.Wrap(err, "failed to make bits dir")
+	}
+
+	// for each of the given file sources
+	for _, file := range b.files {
+
+		fileSrcDst := strings.Split(file, ":")
+
+		if len(fileSrcDst) != 2 {
+			return nil, errors.New("invalid value for --with-file option")
```
fabriziopandini, line 59: what about adding the wrong file name to the error?

kinder/pkg/build/bits/fileBits.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+
+		// Creates an extractor instance, that will read the file bits from src,
+		// and save it to the dest path (inside HostBitsPath)
+		e := extract.NewExtractor(
+			fileSrcDst[0], dst,
```
fabriziopandini, line 65: This will fail if two files have the same name.
My suggestion is to use a tmp name and to maintain in a map[string]string with the mapping from source (or target) names to tmp names

prasadkatti: yeah, good catch. Will update according to your suggestion.

kinder/cmd/kinder/build/nodevariant/nodevariant.go
fabriziopandini: That's an interesting point

Copy paths is more powerful, but as of today the use case for this flag are add a CNI manifest, add a contanerd config file, might be add a static pod patch.

So, unless @neolit123 have strong preferences, I leave to @prasadkatti to evaluate if we can switch to copy path with a reasonable effort within the context of this PR

neolit123: ok, leaving the decision to @prasadkatti

kinder/pkg/build/bits/fileBits.go
prasadkatti: Hey @fabriziopandini. Will I have to update the fileNameMutator in order to be able to give the files a temporary name?

kinder/cmd/kinder/build/nodevariant/nodevariant.go
prasadkatti: I am fairly close to getting `with-file` to work. But I can evaluate how much change is required to get `with-path` to work. If it is not too big of an effort, I can do it just as part of this PR.

kinder/pkg/build/bits/fileBits.go
jiatongw: Destination directory must be an absolute path, right? what if user input a relative path? thoughts on maybe adding a check to parse if the destination directory starts from a "/" or not?

kinder/pkg/build/bits/fileBits.go
```diff
@@ -0,0 +1,105 @@
...
+		files: files,
+	}
+}
+
+// Prepare implements bits.Prepare
+func (b *fileBits) Prepare(c *BuildContext) (map[string]string, error) {
+
+	// ensure the dest path exists on host at HostBitsPath
+	dst := filepath.Join(c.HostBitsPath(), "files")
+	if err := os.Mkdir(dst, 0777); err != nil {
```
jiatongw, line 49: Destination directory must be an absolute path, right? what if user input a relative path? thoughts on maybe adding a check to parse if the destination directory starts from a "/" or not?

kinder/pkg/build/bits/fileBits.go
fabriziopandini: yes, you are right.
The mutator requires a new `nameOverride` field and the `Mutate` func should act accordingly.
It is also required a new option for setting the field while calling `NewExtractor` (something similar to `WithNamePrefix`)

kinder/pkg/build/bits/fileBits.go
prasadkatti: I can add a check to ensure that the destination directory is an absolute path. 👍

kinder/pkg/build/bits/fileBits.go
prasadkatti: Thanks. I was able to get a new `nameOverride` field as you described above. For the original problem of having 2 files with the same name, we can just append the array index to come up with a temporary name.

kinder/pkg/build/bits/fileBits.go
prasadkatti: done

kinder/cmd/kinder/build/nodevariant/nodevariant.go
prasadkatti: Updated the PR to add `with-path` instead of `with-file`.

prasadkatti:
@fabriziopandini @neolit123 - This is ready for another review. Thanks.

kinder/pkg/build/bits/pathBits.go
```diff
@@ -0,0 +1,117 @@
...
+// Install implements bits.Install
+func (b *pathBits) Install(c *BuildContext) error {
+
+	for idx, path := range b.paths {
+		pathSrcDst := strings.Split(path, ":")
+
+		// The src path is a subfolder into the alterDir, that is mounted in the container as /alter
+		src := filepath.Join(c.ContainerBitsPath(), "files", fmt.Sprintf("%s_%d", filepath.Base(pathSrcDst[0]), idx))
+
+		// copy file or directory
```
fabriziopandini, line 101: Might be we should call mkdir to ensure the target directory exists before calling cp

prasadkatti: done

prasadkatti:
One thing I forgot to mention is that the destination path needs to be absolute, but also complete. What I mean by that is if you try -
`kinder build nv --with-path OWNERS:/root/`
the file will get copied to the image inside `/root` with the temporary name (`OWNERS_0`). To work around that you will have to provide the _complete_ destination path -
`kinder build nv --with-path OWNERS:/root/OWNERS`

PR #2215, opened 2020-07-06
kinder: fix wrong patch format in patches-tasks.yaml

neolit123:
The +json patch file should include a patch that follows RFC6902.

otherwise kubeadm complains that it cannot parse the given patch file for the explicitly specified format.
(expected)


neolit123:
/retest

> task-01-add-kubernetes-versions expand_less | 10m0s
> timeout. The task did not complete in less than 10m0s as expected

Issue #2223, opened 2020-07-19
kinder: the latest etcd image cannot be pulled in the containerd base image


neolit123:
the kinder base image (see [1]) we use with containerd for the kinder-regular-* jobs fails to pull the etcd image that was recently updated in https://github.com/kubernetes/kubernetes/pull/92349.

this a appears to be a ctr / containerd problem that i don't understand. possibly related to:
https://github.com/containerd/containerd/issues/3762

> In case anyone else reads this, I faced same issue with containerd 1.3.3 running in kata pod, turned out it was to do with `/var/lib` being on a 9p filesystem, which doesn't support `mknod`. For my use-case It was sufficient for to set `root = /run/containerd/root` and `state = /run/containerd/state` in the containerd config file, as `/run` is a `tmpfs` and I also got some performance improvements on image pulls that way (I do have plenty of memory to spare, at least at the moment).

more details in https://github.com/kubernetes/kubernetes/issues/93223

> ok, looks like there is a ctr / image layer problem:
> 
> ```
> DEBU[21:45:35] Running: [docker exec kind-build-8811a175-3bfe-4b62-80c7-c2c9ce5dd58d bash -c ctr image pull k8s.gcr.io/pause:3.2 > /dev/null] 
> DEBU[21:45:38] Running: [docker exec kind-build-8811a175-3bfe-4b62-80c7-c2c9ce5dd58d ctr image export /kind/images/pause.tar k8s.gcr.io/pause:3.2] 
> DEBU[21:45:38] Running: [docker exec kind-build-8811a175-3bfe-4b62-80c7-c2c9ce5dd58d ctr --namespace=k8s.io image import /kind/images/pause.tar --no-unpack] 
> DEBU[21:45:38] Running: [docker exec kind-build-8811a175-3bfe-4b62-80c7-c2c9ce5dd58d rm /kind/images/pause.tar] 
> DEBU[21:45:39] Running: [docker exec kind-build-8811a175-3bfe-4b62-80c7-c2c9ce5dd58d bash -c ctr image pull k8s.gcr.io/etcd:3.4.9-1 > /dev/null] 
[...]

PR #2228, opened 2020-07-21
kinder: prepull the additional images (pause, etcd..) on the host

neolit123:
Pulling the etcd image in the containerd version of the build-container
can error out for unknown reason with "failed to convert whiteout file".

Instead of pulling images using the container runtime in the
build-container, use the container runtime on the host and copy / import
the images as tarballs.

fixes https://github.com/kubernetes/kubeadm/issues/2223
xref https://github.com/kubernetes/kubernetes/issues/93223

### DIFF
--- a/kinder/pkg/build/alter/alter.go
+++ b/kinder/pkg/build/alter/alter.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"path/filepath"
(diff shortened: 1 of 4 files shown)


neolit123:
i've restored the approved label wit the idea to merge this by EOD today and check if it fixes the signal.
we can get back to the idea and potentially rework this once @fabriziopandini is back.

PR #2242, opened 2020-08-04
verify_manifest_lists.go: fix "panic: could not parse "v1.18.6\n" as version [recovered]" issue

MaXinjian:
verFromURL we get at first is v1.18.6\n, however version.MustParseGeneric can only parse
something like 1.18.6. so we should remove the first and last character

$ curl -s https://storage.googleapis.com/kubernetes-release/release/stable-1.txt
v1.18.6

### DIFF
--- a/tests/e2e/manifests/verify_manifest_lists.go
+++ b/tests/e2e/manifests/verify_manifest_lists.go
@@ -631,6 +631,7 @@ func filterVersions(versions VersionList) (VersionList, error) {
 		if err != nil {
 			return versions, err
 		}
+		verFromURL = strings.TrimSpace(verFromURL)
 		minEstimated := version.MustParseGeneric(verFromURL)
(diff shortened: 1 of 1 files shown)


tests/e2e/manifests/verify_manifest_lists.go
```diff
@@ -631,6 +631,7 @@ func filterVersions(versions VersionList) (VersionList, error) {
 		if err != nil {
 			return versions, err
 		}
+		verFromURL = verFromURL[1:len(verFromURL)-1]
```
neolit123, line 634: > "panic: could not parse "v1.18.6\n" as version [recovered]"

where are you seeing this panic?

better to use:
```
verFromURL = strings.TrimSpace(verFromURL)
```

the leading "v" would be parsed.

MaXinjian: I find this panic when doing #2243 
Thanks and done

Issue #2252, opened 2020-08-13
kinder: tasks do not terminate the e2e test suite after timeout


neolit123:
if a kinder workflow task that is running the e2e test suite timeouts after N minutes, kinder would end this task, but continue running the rest of the tasks and also the suite itself would continue running in the background(?).

https://storage.googleapis.com/k8s-gubernator/triage/index.html?text=error%20loading%20KubeConfig#f44509f17078669e57ea
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-kubeadm-kinder-upgrade-1-19-master/1293358669374689282
https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-e2e-kubeadm-kinder-upgrade-1-19-master/1293358669374689282/build-log.txt

search for: "The task did not complete in less than 25m0s as expected"

what happens there is that kinder would then not skip subsequent tasks if they are using `force: true`:
https://github.com/kubernetes/kubeadm/blob/master/kinder/ci/workflows/upgrade-tasks.yaml#L140

if these are cleanup tasks they would then delete the kubeconfig file for the cluster and the cluster itself and the suite running in the background will start failing.

while the original cause for the problems here is the recent GCE changes that we started seeing (e.g. reduced CPU resources to not complete the suite in <25min), kinder ideally should at least try to terminate the e2e test suite properly so that it does not run in the background.

this appears to be spamming the "k8s-gubernator/triage" board a lot as one of the top "flakes".

neolit123:
PR: https://github.com/kubernetes/kubeadm/pull/2253

neolit123:
one side effect of merging #2253 is that once a process exits naturally it results in:

```
Error: error running test: exit status 1
error: failed obtaining the pgid for pid: 104276, no such process
falling back to killing the parent process only...
error: failed killing process with pid: 104276, os: process already finished
```

a solution would be to check if the process is still running before killing or just return right after the pgid fetch failed.
but overall this is not a big problem.

/remove-priority important-soon
/priority backlog
/reopen

fabriziopandini:
The error are not super intuitive, might be we should wrap them and add a new line that states clearly where the cleanup procedure starts.

PR #2253, opened 2020-08-13
kinder: avoid orphaning child processes on task timeouts

neolit123:
When a task has timed out, ideally the whole process tree should be
terminated with a SIGKILL.

The e2e suite has some caveats with AfterSuite() functions and
these can only contribute noise to the logs. A SIGKILL to
the parent process (Golang command) and the child processes such
as Ginkgo, ensure that everything is terminated right away.

Otherwise, Ginkgo can continue running while the cluster is
being destroyed and fill the logs with errors.

Originally this code was disabled for debugging, but executing it
is mandatory. Add more debug messages to help find problems around
obtaining a pgid or killing a process tree in case failures in CI
are seen again.

fixes https://github.com/kubernetes/kubeadm/issues/2252

### DIFF
--- a/kinder/pkg/test/workflow/taskCmdRunner.go
+++ b/kinder/pkg/test/workflow/taskCmdRunner.go
@@ -287,18 +287,23 @@ func cleanup(cmd *exec.Cmd) {
 		}
 	}()
 
-	/* temporary disabled to better investigate test-grid failures
+	// obtain the process ground ID
 	pgid, err := syscall.Getpgid(cmd.Process.Pid)
 	if err != nil {
-		cmd.Process.Kill()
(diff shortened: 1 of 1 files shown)


neolit123:
/assign @fabriziopandini 
/area kinder
/kind bug
/priority important-soon

if this starts failing in CI again, i will take a look and try to fix it.

neolit123:
> The e2e suite has some caveats with AfterSuite() functions and
these can only contribute noise to the logs. A SIGKILL to
the parent process (Golang command) and the child processes such
as Ginkgo, ensure that everything is terminated right away.

note, in case we are running the e2e suite in a task, a SIGKILL can cause the suite to leave the cluster with leftover objects or partial state. this is not great, but if the suite has failed arguably the main purpose of the test job has failed and we no longer need the cluster and we can proceed to clean it up.

PR #2268, opened 2020-08-31
fix wrong kinder shellcheck url

MaXinjian:
1. change to use stable version
  2. change shellcheck url since old url is unavailable now


fabriziopandini:
/approve

@MaXinjian If possible, group all shellcheck changes in one PR (so it will be easier for reviewers to handle them 😉 )

MaXinjian:
> /approve
> 
> @MaXinjian If possible, group all shellcheck changes in one PR (so it will be easier for reviewers to handle them 😉 )

Yes, I should. But I just found this  shellcheck  issue early this morning.  And I have searched all shellcheck url, seems this is the only wrong url left now.

PR #2269, opened 2020-08-31
fix small shellcheck issue

MaXinjian:
fix the following issue:
    In ./hack/verify-whitespace.sh line 28:
    if [[ ! -z "$TRAILING" ]]; then
          ^-- SC2236: Use -n instead of ! -z.

    For more information:
      https://www.shellcheck.net/wiki/SC2236 -- Use -n instead of ! -z.


fabriziopandini:
/approve

@MaXinjian If possible, group all shellcheck changes in one PR (so it will be easier for reviewers to handle them 😉 )

PR #2297, opened 2020-09-23
kinder: increase e2e test suite timeouts to 35 minutes

neolit123:
It seems 25 minutes is often times not enough for the suite
(parallel runs), either due to more tests being added
or slower infrastructure.

Increase the suite timeout to 35 minutes to try to solve these
problems in the kubeadm/kinder e2e test jobs.

most of the recent flakes in this Summary:
https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-kubeadm
area caused by the 25 minute timeout:

https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-kubeadm#kubeadm-kinder-upgrade-1-19-master

but there are also failures around slow infra that cannot build the kubeadm e2e suite in less than 10 minutes:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-kubeadm-kinder-1-19/1308645363783045120
(10 should be enough, but this will be investigated separatelly)

fixes: https://github.com/kubernetes/kubernetes/issues/90761

### DIFF
--- a/kinder/ci/workflows/external-etcd.yaml
+++ b/kinder/ci/workflows/external-etcd.yaml
@@ -90,7 +90,7 @@ tasks:
     - --parallel
     - --name={{ .vars.clusterName }}
     - --loglevel=debug
-  timeout: 25m
+  timeout: 35m
 - name: get-logs
   description: |
     Collects all the test logs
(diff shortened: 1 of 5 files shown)


yagonobre (commented):
Thanks!
/lgtm

neolit123:
this here is a `timeout. The task did not complete in less than 5m0s as expected` flake, but we should address separately.

PR #2314, opened 2020-09-30
kinder: increase the timeout for kubeadm e2e suite on presubmit

neolit123:
The default timeout of 5m causes frequent timeouts.

for example:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubeadm/2313/pull-kubeadm-kinder-upgrade-latest/1311100923589693440


fabriziopandini (commented):
/approve
/lgtm

Issue #2317, opened 2020-09-30
[Roadmap] Improve kubeadm support for declarative approaches/git-ops


fabriziopandini:
Kubeadm, being a CLI, does not play well with declarative approaches/git-ops workflows.

Assuming that kubeadm is divided in two main parts
1. bootstrapping a node (transforming a machine into a node: init, join)
2. managing an existing node (e.g. upgrades, renew certs, changing a node)

This issue is about collecting ideas and define a viable path for making 2 possible using declarative approaches, sometimes referred also as in-place mutations. 

For this first iteration, I consider 1 out of scope, mainly because bootstrapping nodes with a declarative approach is already covered by Cluster API and it is clearly out of the scope of kubeadm.

fabriziopandini:
Prior discussion from https://github.com/kubernetes/kubeadm/issues/1698

@timothysc 

> As a Kubernetes Operator I would like to enable be able to declaratively control configuration changes, and upgrades in a systematic fashion.

@fabriziopandini 

> IMO the kubeadm operator should be responsible for two things
> - In place mutations of kubeadm generated artifacts
> - Orchestration of such mutations across nodes
> Instead, I think that we should consider out of scope everything that fits under the management of infrastructure or it is related to the management of "immutable" nodes (where "Immutable" = any operation done deploying a new node and removing the old one)

@neolit123 

> my other top question, this can end up being not-so-secure.

fabriziopandini:
For the kubeadm operator, I think we should focus on the first use case, "declaratively control configuration changes", given that this is not supported by kubeadm now and it was a top priority in the recent survey

fabriziopandini:
Draft KEP https://docs.google.com/document/d/14Cb2fQfRVpPSQuNz0MYbtGmO63xBglT7x7h57ZeK5PI/edit?usp=sharing

jhughes2112:
Interesting proposal.  What I've struggled with over the past year using kubeadm is specifically what I addressed with my own scripts that (procedurally) builds clusters (on github: [k8smaker](https://github.com/jhughes2112/k8smaker)).  The best practices for configuring a bare metal cluster is pretty complex.  Doing so with AWS is too.  The preconditioning script depends strongly on the underlying OS involved.  But having built it modularly, I can see how the cluster construction (init, join) can be made completely extensible while providing a fully declarative interface to the user.

It requires:
- ssh credentials to access any new node with sudo privileges (from whence kubeadm or the proposed operator executes)
[...]

PR #2333, opened 2020-10-23
delete go.mod in root directory

hwdef:
Fix #1813  again

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,0 @@
-module k8s.io/kubeadm
-
-go 1.12


fabriziopandini:
The go-mod file in the root directory gets automatically generate when updating go modules under `kinder/` or `operator/`, so IMO there is no point in continuing to delete it
I'm +1 for closing this PR

hwdef:
Maybe you can find a way to not generate it.

fabriziopandini:
Feel free to open an issue and propose other solutions, but please take care such solutions are not adding any burden to people working in this repo.
In the meantime, I don't see any problem in keeping the file

hwdef:
fine.

just close this.

hwdef:
@fabriziopandini 
What commond did you used?
I try to update go modules under `kinder/` or `operator/`, but go.mod in root directory is not recreate.

hwdef:
/assign @neolit123

Can you give me some suggestions?
Or we can add go.mod to .gitignore.

fabriziopandini:
TBH, it will be really weird for me to have go.mod in .gitignore

As explained before, as a maintainer of this project, I don't see problems in having this file, and I don't want this to get into the way of contributors again given that after the first PR got merged it was regenerated by simply working on kinder/on the operator.

Also, as soon as the details about the move of kubeadm out of K/K will be addressed, this file will be required, so any effort to remove it is just temporary.

I'm -1 to get this merged
🤷‍♂️

hwdef:
It is wrong to keep files just because there are no problems, and it will cause more and more garbage.
I looked at the file history, this file was submitted by you, and it is not surprising that you object to cleaning it.
Maybe others do not agree to clean up this file, but I still think that you uploaded this file as a mistake.
You can +1 or -1, but please don’t promote bad habits.

hwdef:
@fabriziopandini 
And please let me ask again, what command did you use to upgrade? Did you make a mistake that caused this file to be generated?

fabriziopandini:
> what command did you use to upgrade? Did you make a mistake that caused this file to be generated?

The file was generated working on the operator, but TBH I don't remember if this happened while updating some reference or by running some hack/verify script.
However this is going to be revised as per https://github.com/kubernetes/kubeadm/issues/2317, and this will revamp most of the initial experiment + update to the latest version of golang, so might be the problem will be fixed during this effort

neolit123:
The file is not needed in general since the root folder is not a module. I
can confirm that something generates it under certain conditions. Instead
of this pr if you can find what of our scripts generates it and change that
behavior it would be the better fix.

fabriziopandini:
I just gave another try at this while working on kinder and immediatebly stumbled into
```
package k8s.io/kubeadm: no Go files in /Users/fpandini/go/src/k8s.io/kubeadm
```
when doing `go install` from the kinder folder

PR #2341, opened 2020-11-07
Update some kind depencies

fabriziopandini:
kinder is lagging behind with its own dependencies, so let's start by updating the easy ones.

Also, giving another try to removing go.mod from the root given that it seems that the latest version of go/visual studio code/goland are not regenerating anymore this file automatically.

/assign @neolit123


neolit123:
/area kinder
/lgtm
thanks

PR #2345, opened 2020-11-09
Remove useless cleanup code since src dir never existed

MaXinjian:


fabriziopandini:
/ok-to-test
cc @neolit123

neolit123 (commented):
It existed before go mod was used.
/lgtm
/approve

Issue #2344, opened 2020-11-07
Drop kustomize dependency in kinder


fabriziopandini:
Kustomize dependency in kinder is difficult to upgrade given the indirect dependency with Kubernetes libraries that are reference directly by kinder or indirectly by kind.

Given that kustomize is used only in a single point in the kinder codebase (pkg/kubeadm), I'm proposing to replace it with standard json patches to greatly simplify our dependency tree.

/cc @neolit123

neolit123:
we could replace with the "patches" package from k/k/cmd/kubeadm if that is moved to a the kubeadm library.

/area kinder
/kind design cleanup
/priority backlog

PR #2350, opened 2020-11-14
Remove Kustomize dependency and use json patch instead

fabriziopandini:
This PR removes Kustomize dependency and uses json patch instead, so the resulting dependency tree is simplified.

fixes: https://github.com/kubernetes/kubeadm/issues/2344

/assign @neolit123


kinder/pkg/kubeadm/patch.go
```diff
@@ -0,0 +1,298 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
neolit123, line 2: 2020

kinder/pkg/kubeadm/patch.go
```diff
@@ -0,0 +1,298 @@
...
+	"io"
+	"strings"
+
+	jsonpatch "github.com/evanphx/json-patch"
+	"github.com/pkg/errors"
+	"sigs.k8s.io/yaml"
+)
+
+/*
+patch.go provide utilities for applying patches to a YAML file.
```
neolit123, line 31: provides

kinder/pkg/kubeadm/patch.go
```diff
@@ -0,0 +1,298 @@
...
+	Version string `json:"version"`
+	Kind    string `json:"kind"`
+	// Patch should contain the contents of the json patch as a string
+	Patch string `json:"patch"`
+}
+
+// Build takes a Kubernetes object YAML document stream to patch,
+// merge patches, and JSON 6902 patches.
+//
+// It returns a patched a YAML document stream.
```
neolit123, line 51: a YAML -> YAML

neolit123 (commented):
@fabriziopandini  thanks.
found a couple of typos.

instead of forking kind's patch management we could have waited for exposing https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/util/patches/patches.go in the kubeadm-library.

but this could be done in another stage on top of this.

neolit123 (commented):
/lgtm

PR #2377, opened 2021-01-18
tests/e2e/manifests: adapt to recent coredns image change

neolit123:
new coredns image is located at a coredns subpath.

fixes these failures:
https://testgrid.k8s.io/sig-cluster-lifecycle-all#periodic-manifest-lists


neolit123:
self-merging to not bother people with this small change.

Issue #2381, opened 2021-01-27
"kinder delete cluster" is not working


neolit123:
steps to reproduce:
```
kinder create cluster --name foo
kinder delete cluster --name foo
```

it says `Deleting "foo"...` but the containers (nodes) remain running which can be confirmed with `docker ps -a`

PR #2383, opened 2021-01-30
kinder: sync ClusterLabelKey and NodeRoleLabelKey with kind

SataQiu:
Kind has migrated to use the new label `io.x-k8s.kind.cluster` and `io.x-k8s.kind.role`.
But Kinder is still using the old labels, this is why the cluster deletion failed.
This PR marks the old labels as DEPRECATED and starts to use the new labels.

Fixes: #2381
Ref: https://github.com/kubernetes-sigs/kind/pull/1074

https://github.com/kubernetes-sigs/kind/commit/a39a4efc8475817f1117026a3259d89524eafc28#diff-980007e783aef4dcb904e3684c6cff1b43a651cbd0134db24f4a45f331adc5fc

### DIFF
--- a/kinder/pkg/cluster/status/cluster.go
+++ b/kinder/pkg/cluster/status/cluster.go
@@ -72,9 +72,9 @@ func ListClusters() ([]string, error) {
 		"-a",         // show stopped nodes
 		"--no-trunc", // don't truncate
 		// filter for nodes with the cluster label
-		"--filter", "label="+constants.ClusterLabelKey,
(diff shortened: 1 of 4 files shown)


neolit123 (commented):
thanks for the fix.

/lgtm
/approve

neolit123:
> neolit123 added area/kinder kind/bug and removed approved lgtm labels yesterday

hm, odd
re-adding the labels.

PR #2396, opened 2021-02-24
kinder: add kubelet skew test jobs

neolit123:
on the SIG Node / Arch mailing list there is a discussion around the fact that the kubelet claims support for N-2 skew against the kube-apiserver but currently there are not tests for that.

https://groups.google.com/d/msgid/kubernetes-sig-architecture/CAH1uJ6U4qftRnmWthjYKtBKAzRFFnX7XWLMkSbyVWHUsX8%3DBBg%40mail.gmail.com?utm_medium=email&utm_source=footer

on the side of kubeadm we always recommend that users match their kubelet version with the control plane version and the kubeadm version, but such skew tests can help SIG Node and feature upgrade testing in core.

these commit add tests for version skew testing with kubeadm / kinder:
```
skew-kubelet-1.17-on-1.18.yaml
skew-kubelet-1.17-on-1.19.yaml
skew-kubelet-1.18-on-1.19.yaml
skew-kubelet-1.18-on-1.20.yaml
skew-kubelet-1.19-on-1.20.yaml
skew-kubelet-1.19-on-latest.yaml
skew-kubelet-1.20-on-latest.yaml
```

NOTE: x-on-y jobs that y==x are already covered by the `regular*` jobs.

one caveat (as the note in kubeadm-periodic.tests.md explains) is that kubeadm doesn't really support N-2 kubelet skew (max is N-1), so in the future if changes to the kubelet happen such as these tests are no longer possible with kubeadm directly we need to re-evaluate.

### DIFF
--- a/kinder/ci/workflows/skew-kubelet-1.17-on-1.18.yaml
+++ b/kinder/ci/workflows/skew-kubelet-1.17-on-1.18.yaml
@@ -1,0 +1,14 @@
+version: 1
+summary: |
+  This workflow tests the proper functioning of kubeadm and control plane from 1.18 against 1.17 kubelet
+  test grid > https://testgrid.k8s.io/sig-cluster-lifecycle-kubeadm#kubeadm-kinder-kubelet-1.17-on-1.18
+  config    > https://git.k8s.io/test-infra/config/jobs/kubernetes/sig-cluster-lifecycle/kubeadm-kinder-kubelet-x-on-y.yaml
+  config    > https://git.k8s.io/test-infra/testgrid/config.yaml
+vars:
+  kubeadmVersion: "{{ resolve `ci/latest-1.18` }}"
+  kubeletVersion: "{{ resolve `ci/latest-1.17` }}"
+  kubernetesVersion: "{{ resolve `ci/latest-1.18` }}"
+  ignorePreflightErrors: "KubeletVersion"
+  controlPlaneNodes: 3
+tasks:
+- import: skew-x-on-y-tasks.yaml
--- a/kinder/ci/workflows/skew-x-on-y-tasks.yaml
+++ b/kinder/ci/workflows/skew-x-on-y-tasks.yaml
@@ -7,6 +7,7 @@ vars:
   # vars defines default values for variable used by tasks in this workflow;
   # those values might be overridden when importing this files.
   kubeadmVersion: v1.13.5
+  kubeletVersion: v1.12.8
   kubernetesVersion: v1.12.8
   controlPlaneNodes: 1
   workerNodes: 2
@@ -14,6 +15,8 @@ vars:
   image: kindest/node:test
   clusterName: kinder-xony
   kubeadmVerbosity: 6
+  defaultIgnorePreflightErrors: Swap,SystemVerification,FileContent--proc-sys-net-bridge-bridge-nf-call-iptables,
+  ignorePreflightErrors: ""
 tasks:
 - name: pull-base-image
   description: |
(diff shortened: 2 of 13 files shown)


neolit123:
/hold

@fabriziopandini leaving this up for discussion for now.
PTAL and LMK what you think.

kinder/ci/workflows/skew-x-on-y-tasks.yaml
```diff
@@ -7,35 +7,39 @@ vars:
...
-  args:
-    - build
-    - node-image-variant
-    - --base-image={{ .vars.baseImage }}
-    - --image={{ .vars.image }}
-    - --with-init-artifacts={{ .vars.kubernetesVersion }}
-    - --with-kubeadm={{ .vars.kubeadmVersion }}
-    - --loglevel=debug
-  timeout: 15m
+# - name: pull-base-image
```
fabriziopandini, line 21: left-over from local testing?

kinder/ci/workflows/skew-kubelet-1.17-on-1.18.yaml
```diff
@@ -0,0 +1,14 @@
...
+summary: |
+  This workflow tests the proper functioning of kubeadm and control plane from 1.18 against 1.17 kubelet
+  test grid > https://testgrid.k8s.io/sig-cluster-lifecycle-kubeadm#kubeadm-kinder-kubelet-1.17-on-1.18
+  config    > https://git.k8s.io/test-infra/config/jobs/kubernetes/sig-cluster-lifecycle/kubeadm-kinder-kubelet-X-on-Y.yaml
+  config    > https://git.k8s.io/test-infra/testgrid/config.yaml
+vars:
+  kubeadmVersion: "{{ resolve `ci/latest-1.18` }}"
+  kubeletVersion: "{{ resolve `ci/latest-1.17` }}"
+  kubernetesVersion: "{{ resolve `ci/latest-1.18` }}"
+  ignorePreflightErrors: ",KubeletVersion"
```
fabriziopandini, line 11:
```suggestion
  ignorePreflightErrors: "KubeletVersion"
```
Let's move the responsibility of adding the initial comma to the template in skew-x-on-y-tasks.yaml

kinder/ci/workflows/skew-x-on-y-tasks.yaml
```diff
@@ -71,75 +76,76 @@ tasks:
...
-  description: |
-    Deletes the cluster
-  cmd: kinder
-  args:
-    - delete
-    - cluster
-    - --name={{ .vars.clusterName }}
-    - --loglevel=debug
-  force: true
+# - name: cluster-info
```
fabriziopandini, line 83: same

fabriziopandini (commented):
I'm not opposed to having these test in the kubeadm repo, however, there should be some kind of agreement on the responsibility for monitoring these job/triage failures (I guess the first line here is SIG Node, with our team at support)

kinder/ci/workflows/skew-x-on-y-tasks.yaml
neolit123: indeed, i blame the late hours working on this.

kinder/ci/workflows/skew-kubelet-1.17-on-1.18.yaml
neolit123: i need to check if this is a problem for the trailing `,` for jobs that don't pass the extra `ignorePreflightErrors`.
this may be solvable with go template `if / else` but i don't like it because it's a bit messy.

neolit123:
> I'm not opposed to having these test in the kubeadm repo, however, there should be some kind of agreement on the responsibility for monitoring these job/triage failures (I guess the first line here is SIG Node, with our team at support)

i proposed that we duplicate / add the jobs to a SIG Node dashboard.

kinder/ci/workflows/skew-kubelet-1.17-on-1.18.yaml
neolit123: tested with the kubeadm CLI and a trailing `,` seems to be ignored.

neolit123:
@fabriziopandini updated.

neolit123:
let's iterate if we find problems.

/hold cancel

PR #2417, opened 2021-03-22
verify_manifest_lists: don't test conformance:v1.21.0-beta.1

neolit123:
This image is missing in the pre-release, so we have to skip it.
Instead of skipping images like that ideally the release process
should release again this missing image, but looks like this is
complicated.

discussion:
https://kubernetes.slack.com/archives/C2C40FMNF/p1615817136017300?thread_ts=1615817136.017300&cid=C2C40FMNF

fixes https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-all#periodic-manifest-lists


fabriziopandini (commented):
:-(
/lgtm

PR #2421, opened 2021-03-26
change config.yaml instead of args for fail-swap-on

pacoxu:
> Mar 26 02:30:25 kind-control-plane kubelet[244]: Flag --fail-swap-on has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/100573/pull-kubernetes-e2e-kind-ipv6/1375269122421559296/artifacts/logs/kind-control-plane/kubelet.log

Fix warning logs in e2e testing envs.

### DIFF
--- a/kinder/pkg/build/bits/initBits.go
+++ b/kinder/pkg/build/bits/initBits.go
@@ -278,14 +278,6 @@ func configureKubelet(c *BuildContext) error {
 		return err
 	}
 
-	// ensure we don't fail if swap is enabled on the host
-	if err := c.RunInContainer("/bin/sh", "-c",
-		`echo "KUBELET_EXTRA_ARGS=--fail-swap-on=false" >> /etc/default/kubelet`,
-	); err != nil {
-		log.Errorf("failed to set kubelet fail-swap-on! %v", err)
-		return err
-	}
-
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


kinder/pkg/build/bits/initBits.go
```diff
@@ -280,7 +280,7 @@ func configureKubelet(c *BuildContext) error {
 
 	// ensure we don't fail if swap is enabled on the host
 	if err := c.RunInContainer("/bin/sh", "-c",
-		`echo "KUBELET_EXTRA_ARGS=--fail-swap-on=false" >> /etc/default/kubelet`,
+		`echo "FailSwapOn: false" >> /var/lib/kubelet/config.yaml`,
```
neolit123, line 283: kubeadm init and join regenerate this file and it will be overwritten.

the /etc/default/kubelet override does works but it's for flags only.

i think the value should be added here instead:
https://github.com/kubernetes/kubeadm/blob/05f9b96cf83d73ed2177815487eba90517b66708/kinder/pkg/kubeadm/config.go#L200-L216

neolit123 (commented):
/lgtm
/approve

if the CI test here passes then the kubelet is receiving the `failSwapOn: false` properly.

Issue #2420, opened 2021-03-25
kinder: switch to a newer CNI


neolit123:
we pinned the Calico (CNI) version to avoid downloading the latest version.
our old version of calico is using deprecated APIs (i think it was CRD v1beta1 and it will be removed in 1.22), so we should switch to the latest.

the latest Calico install includes an operator and have two manifests:
```
kubectl create -f https://docs.projectcalico.org/manifests/tigera-operator.yaml
kubectl create -f https://docs.projectcalico.org/manifests/custom-resources.yaml
```
https://docs.projectcalico.org/getting-started/kubernetes/quickstart

the last time i checked this it did not work for me, maybe because i was running a single CP node.
we are currently pinning calico:
https://github.com/kubernetes/kubeadm/blob/05f9b96cf83d73ed2177815487eba90517b66708/kinder/pkg/data/calicomanifests.go

we can switch to Cillium:
```
kubectl create -f https://raw.githubusercontent.com/cilium/cilium/v1.9/install/kubernetes/quick-install.yaml
```

this would allow us to better pin the CNI version and worked fine for me.

fabriziopandini:
Why not switching to kindness? we are using it in CAPD now https://github.com/kubernetes-sigs/cluster-api/blob/master/test/e2e/data/cni/kindnet/kindnet.yaml

neolit123:
I have not tried it. Is it production ready?

One downside is that it is not something that users commonly use in
production. So i cannot tell them 'Try kindnet. We know it works in our CI'.

neolit123:
got Calico working, but only with ClusterConfiguration->networking->podSubnet: 192.168.0.0/16
Calico at one point supported custom CIDR allocation and this was not needed...

i'm realizing that CNI plugins have become *huge* and slow to start due to their operators and other bits:
```
quay.io/tigera/operator              v1.15.1              73b255331773        2 weeks ago         47.5MB
calico/node                          v3.18.1              50b52cdadbcf        2 weeks ago         172MB
calico/typha                         v3.18.1              cdfbea6501a9        2 weeks ago         52.8MB
calico/pod2daemon-flexvol            v3.18.1              3994c62982cc        2 weeks ago         21.7MB
calico/cni                           v3.18.1              21fdaa2fccee        2 weeks ago         131MB
calico/kube-controllers              v3.18.1              a274b7fd57f5        2 weeks ago         53.4MB
```

```
[...]

PR #2424, opened 2021-03-28
kinder: using kindnet CNI

SataQiu:
Fix: #2420

As `kindnet` is the default CNI for `kind` cluster and `Calico` CNI plugins have become huge and slow to start, we prefer to use `kindnet` as `kinder` CNI plugin.


kinder/pkg/cluster/manager/actions/kubeadm-config.go
```diff
@@ -92,8 +92,8 @@ func KubeadmConfig(c *status.Cluster, kubeDNS bool, copyCertsMode CopyCertsMode,
 		APIBindPort:          constants.APIServerPort,
 		APIServerAddress:     controlPlaneIP,
 		Token:                constants.Token,
-		PodSubnet:            "192.168.0.0/16", // default for calico
-		ServiceSubnet:        "",               // let kubeadm apply default
+		PodSubnet:            "192.168.0.0/16",
```
neolit123, line 95: oddly this failed for me locally...or maybe i'm mistaken because the CI test passed on this PR.

kinder/pkg/cluster/manager/actions/kubeadm-config.go
```diff
@@ -92,8 +92,8 @@ func KubeadmConfig(c *status.Cluster, kubeDNS bool, copyCertsMode CopyCertsMode,
...
+		ServiceSubnet:        "", // let kubeadm apply default
```
neolit123, line 96: we can remove this too?

kinder/pkg/data/kindnetmanifests.go
```diff
@@ -0,0 +1,134 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
neolit123, line 2: 2021

neolit123 (commented):
thanks @SataQiu 
added a couple of minor comments.

we should also have this PR as `Fixes #2420`

EDIT: actually we need to update https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm/#pod-network

SataQiu:
> thanks @SataQiu
> added a couple of minor comments.
> 
> we should also have this PR as `Fixes #2420`
> 
> EDIT: actually we need to update https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm/#pod-network

How about just removing the node about Calico CNI? 
Whether the CNI can work properly seems to be beyond kubeadm's scope.

```
Note: Currently Calico is the only CNI plugin that the kubeadm project performs e2e tests against. If you find an issue related to a CNI plugin you should log a ticket in its respective issue tracker instead of the kubeadm or kubernetes issue trackers.
```

kinder/pkg/cluster/manager/actions/kubeadm-config.go
SataQiu: It looks like `PodSubnet` is also required by `kindnet`.
https://github.com/SataQiu/kindnet/blob/d3c882ff20054f504e3bb7cf3637c74cf9a332e4/cmd/kindnetd/subnets.go#L65-L72

kinder/doc/kind-kinder.md
```diff
@@ -56,7 +56,6 @@ _Actions on a running cluster:_
         - shortcut for testing different kubeadm join discovery mechanics
     - `kubeadm init` can be executed as a unique workflow or using phases
     - `kubeadm join` can be executed as a unique workflow or using phases
-    - the init action installs Calico as a CNI plugin instead of kindnet
```
fabriziopandini, line 59: Let's keep a note we are using kindnet, which is acceptable given that kubeadm should be CNI agnostic and validation of CNI providers is out of scope of our current E2E testing

kinder/pkg/data/kindnetmanifests.go
```diff
@@ -0,0 +1,134 @@
...
+            - name: HOST_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.hostIP
+            - name: POD_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
+            - name: POD_SUBNET
+              value: "192.168.0.0/16"
```
SataQiu, line 102: POD_SUBNET is defined here. @neolit123

kinder/pkg/data/kindnetmanifests.go
```diff
@@ -0,0 +1,134 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package data
+
+// KindnetCNI0_5_4 is the manifest for kindnet 0.5.4
+const KindnetCNI0_5_4 = `
```
fabriziopandini, line 20: I know this could be a stretch for this PR, but what about using the new go v1.16 embed package instead of embedding yaml strings in our codebase?

SataQiu: Sounds good. It's cool!

kinder/pkg/cluster/manager/actions/kubeadm-init.go
```diff
@@ -32,7 +33,11 @@ import (
 
 	"k8s.io/kubeadm/kinder/pkg/cluster/status"
 	"k8s.io/kubeadm/kinder/pkg/constants"
-	"k8s.io/kubeadm/kinder/pkg/data"
+)
+
+var (
+	//go:embed assets/kindnet.yaml
+	kindnetManifest string
```
neolit123, line 40: this is nice.

neolit123 (commented):
/lgtm
/hold cancel

neolit123:
but as noted in https://github.com/kubernetes/kubeadm/issues/2420
we are in test freeze. i don't see high risk in merging that, but we should do it when 1.21 opens, if we want to follow the rules.

pacoxu:
/hold cancel
as 1.21 was released last week.

PR #2425, opened 2021-03-30
Indexing may panic because of 'nil' slice

Akiros001:


kinder/pkg/cluster/manager/manage.go
```diff
@@ -151,9 +151,11 @@ func (c *ClusterManager) CopyFile(source, target string) error {
...
 
 	if targetNodes == nil {
-		fmt.Printf("Copying from %s ...\n", sourceNodes[0].Name())
-		if err := sourceNodes[0].CopyFrom(sourcePath, targetPath); err != nil {
-			return err
+		if sourceNodes != nil {
+			fmt.Printf("Copying from %s ...\n", sourceNodes[0].Name())
+			if err := sourceNodes[0].CopyFrom(sourcePath, targetPath); err != nil {
+				return err
+			}
```
neolit123, line 158: we have a `sourceNodes != nil` check above, so i don't think this is exactly needed.

note: you need sign the CLA to contribute to k8s projects.

pacoxu:
Can you reproduce the problem and provide the log?

kinder/pkg/cluster/manager/manage.go
neolit123: the logic here is the following in pseudo code:

```
if sourceNodes == nil && targetNodes == nil { return error }
if sourceNodes != nil && targetNodes != nil { return error }
```

so at this point either sourceNodes or targetNodes is nil, but no both.

/close

Issue #2441, opened 2021-04-13
kinder build at CI is failing


neolit123:
```
Building kinder...
go: downloading github.com/sirupsen/logrus v1.7.0
go: downloading github.com/pkg/errors v0.9.1
go: downloading github.com/spf13/cobra v1.1.1
go: downloading github.com/google/uuid v1.1.2
go: downloading sigs.k8s.io/yaml v1.2.0
go: downloading sigs.k8s.io/kind v0.9.0
go: downloading k8s.io/apimachinery v0.19.4
go: downloading k8s.io/client-go v0.19.4
go: downloading github.com/spf13/pflag v1.0.5
go: downloading gopkg.in/yaml.v2 v2.3.0
go: downloading github.com/evanphx/json-patch v4.9.0+incompatible
go: downloading golang.org/x/sys v0.0.0-20201107080550-4d91cf3a1aaf
go: downloading k8s.io/klog v1.0.0
go: downloading k8s.io/klog/v2 v2.2.0
go: downloading github.com/mattn/go-isatty v0.0.12
go: downloading gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
go: downloading github.com/alessio/shellescape v1.2.2
go: downloading github.com/pelletier/go-toml v1.8.0
go: downloading github.com/BurntSushi/toml v0.3.1
go: downloading github.com/evanphx/json-patch/v5 v5.1.0
go: downloading github.com/go-logr/logr v0.2.0
go: downloading golang.org/x/crypto v0.0.0-20200622213623-75b288015ac9
go: downloading github.com/imdario/mergo v0.3.11
go: downloading golang.org/x/net v0.0.0-20200822124328-c89045814202
go: downloading sigs.k8s.io/structured-merge-diff/v4 v4.0.1
go: downloading github.com/gogo/protobuf v1.3.1
go: downloading golang.org/x/oauth2 v0.0.0-20200902213428-5d25da1a8d43
go: downloading github.com/davecgh/go-spew v1.1.1
go: downloading github.com/google/gofuzz v1.1.0
go: downloading k8s.io/utils v0.0.0-20200729134348-d5654de09c73
go: downloading golang.org/x/time v0.0.0-20191024005414-555d28b269f0
go: downloading github.com/modern-go/reflect2 v1.0.1
go: downloading gopkg.in/inf.v0 v0.9.1
go: downloading github.com/json-iterator/go v1.1.10
[...]

PR #2442, opened 2021-04-13
kinder: import the kindnet manifest as source code

neolit123:
In CI we use a versioned kubekins image that does not
have go1.16 when testing k8s versions older than 1.21.
This means we have to preserve backwards compatibility
and cannot use "go:embed" until all kubekins versions
in the k8s skew have at least go1.16.

fixes https://github.com/kubernetes/kubeadm/issues/2441

### DIFF
--- a/kinder/pkg/cluster/manager/actions/assets/kindnet.go
+++ b/kinder/pkg/cluster/manager/actions/assets/kindnet.go
@@ -1,3 +1,23 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


kinder/pkg/cluster/manager/actions/assets/kindnet.go
```diff
@@ -110,4 +130,5 @@ spec:
             type: FileOrCreate
         - name: lib-modules
           hostPath:
-            path: /lib/modules
+			path: /lib/modules
```
fabriziopandini, line 133: It seems indentation is not correct here

neolit123: fixed.

neolit123:
adding LGTM label to unblock the tests.

Issue #2444, opened 2021-04-14
kubeadm-kinder-no-addons-latest is failing


neolit123:
https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-kubeadm#kubeadm-kinder-no-addons-latest

the test does the following trying to ensure that join and upgrade work without addons:
- calls init
- deletes the kube-proxy and coredns configmaps
- calls join
- checks if the joining node is ready

at this point the test fails, since something changed in the kubelet and the kube-proxy pod on that node never becomes ready.
prior to that, a pod in a DS that mounted an non-existing CM was tolerated, but now it fails.

- ... continues with upgrade

deleting the kube-proxy DS means the node will never be ready unless we use something that provides what kube-poxy provides (e.g. not kindnet, but a CNI that has a proxy)

one solution is to instead of just deleting the kube-proxy CM is to rename it and patch the DS to point to it.
this feels like the easiest patch.

neolit123:
this is fixed https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-kubeadm#kubeadm-kinder-no-addons-latest

PR #2445, opened 2021-04-14
kinder: fix a problem in the upgrade test without addons

neolit123:
The test is failing since a missing kube-proxy config map,
now causes the joining node to never become ready.

Rename the kube-proxy CM to ensure a scenario where
the user has a "custom proxy".

Also during upgrade ignore only the CoreDNS* preflight checks.

more details here:
xref https://github.com/kubernetes/kubeadm/issues/2444


neolit123:
this seems to work locally for me.
self-merging the change and babysitting the signal further if needed.

neolit123:
needs https://github.com/kubernetes/kubeadm/pull/2446 to merge first.

PR #2448, opened 2021-04-15
delete redundant control characters

zhangshj-inspur:
delete redundant control characters


pacoxu:
Since it looks the same to me 😓, I suggest to check all readme files per https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#trivial-edits/.

zhangshj-inspur:
> Since it looks the same to me , I suggest checking all readme files per https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#trivial-edits/.

I delete several control characters,which maybe invisible on your device,you can see it on notepadd++,also I have checked all markdown file in the project and didn't find other file that have this problem.｡◕‿◕｡

neolit123:
/lgtm
/approve
/ok-to-test
/kind cleanup

the existing characters are odd and i don't know where they came from.

Issue #2249, opened 2020-08-06
Kinder: add flag for forcing the kubeadm config version


fabriziopandini:
As discussed in https://github.com/kubernetes/kubeadm/pull/2248#discussion_r466457798, due to external reasons, we are removing "v1beta1" from the kubeadm test grid is kind of worrying because it is still used in CAPI.

We should consider adding a new flag to kinder for forcing the config version to use and add dedicated tests for this (v1.16 with v1beta1, v1.17 with v1beta1 etc)

neolit123:
overall i'm -1 for this.

i really wish for us to stay on schedule for removal of deprecated v1beta1 API in 1.20.
wouldn't CAPI e2e provide signal for it if CAPI still wishes to use it?

neolit123:
ok, maybe we can just shift the 1.16 and 1.17 kinder jobs to use this flag, but leave 1.18 and newer use v1beta2?

fabriziopandini:
Yep. this should not affect out schedule for v1beta1 deprecation
My intention is only to have test coverage about the fact that the older version are actually capable to start a cluster given a v1beta1 config (and there are no regressions on this capability)

Issue #2428, opened 2021-04-01
kinder: update to support v1beta3 and remove v1beta1


neolit123:
with the addition of kubeadm/v1beta3 and removal of v1beta1 we should update kinder to test v1beta3 by default.
https://github.com/kubernetes/kubeadm/search?q=v1beta1

we might also want to action on:
https://github.com/kubernetes/kubeadm/issues/2249
and add one extra job to test kinder with v1beta2 and "regular" cluster creation.

proposal:
- drop v1beta1
- enable v1beta3 by default
- pass the [new flag for api version](https://github.com/kubernetes/kubeadm/pull/2483) with v1beta2 for all jobs that test kubeadm versions older than 1.22

PR #2483, opened 2021-05-21
kinder: add flag for forcing the kubeadm config version

SataQiu:
kinder: add flag for forcing the kubeadm config version

Ref: #2428 
Fixes: #2249

### DIFF
--- a/kinder/cmd/kinder/do/do.go
+++ b/kinder/cmd/kinder/do/do.go
@@ -43,6 +43,7 @@ type flagpole struct {
 	PatchesDir            string
 	Wait                  time.Duration
 	IgnorePreflightErrors string
+	KubeadmConfigVersion  string
 }
 
 // NewCommand returns a new cobra.Command for exec
@@ -120,6 +121,13 @@ func NewCommand() *cobra.Command {
 		"ignore-preflight-errors", constants.KubeadmIgnorePreflightErrors,
 		"list of kubeadm preflight errors to skip",
 	)
+	cmd.Flags().StringVar(
+		&flags.KubeadmConfigVersion,
+		"kubeadm-config-version", flags.KubeadmConfigVersion,
+		"the kubeadm config version to be used for init, join and upgrade. "+
+			"If not set, kubeadm will automatically choose the kubeadm config version "+
+			"according to the Kubernetes version in use",
+	)
 	return cmd
 }
 
@@ -173,6 +181,7 @@ func runE(flags *flagpole, cmd *cobra.Command, args []string) (err error) {
 		actions.VLevel(flags.VLevel),
 		actions.PatchesDir(flags.PatchesDir),
 		actions.IgnorePreflightErrors(flags.IgnorePreflightErrors),
+		actions.KubeadmConfigVersion(flags.KubeadmConfigVersion),
 	)
 	if err != nil {
 		return errors.Wrapf(err, "failed to exec action %s", action)
--- a/kinder/pkg/cluster/manager/actions/actions.go
+++ b/kinder/pkg/cluster/manager/actions/actions.go
@@ -39,13 +39,13 @@ var actionRegistry = map[string]func(*status.Cluster, *RunOptions) error{
 	"kubeadm-config": func(c *status.Cluster, flags *RunOptions) error {
 		// Nb. this action is invoked automatically at kubeadm init/join time, but it is possible
 		// to invoke it separately as well
-		return KubeadmConfig(c, flags.kubeDNS, flags.copyCertsMode, flags.discoveryMode, c.K8sNodes().EligibleForActions()...)
+		return KubeadmConfig(c, flags.kubeadmConfigVersion, flags.kubeDNS, flags.copyCertsMode, flags.discoveryMode, c.K8sNodes().EligibleForActions()...)
 	},
 	"kubeadm-init": func(c *status.Cluster, flags *RunOptions) error {
-		return KubeadmInit(c, flags.usePhases, flags.kubeDNS, flags.copyCertsMode, flags.patchesDir, flags.ignorePreflightErrors, flags.wait, flags.vLevel)
+		return KubeadmInit(c, flags.usePhases, flags.kubeDNS, flags.copyCertsMode, flags.kubeadmConfigVersion, flags.patchesDir, flags.ignorePreflightErrors, flags.wait, flags.vLevel)
 	},
 	"kubeadm-join": func(c *status.Cluster, flags *RunOptions) error {
-		return KubeadmJoin(c, flags.usePhases, flags.copyCertsMode, flags.discoveryMode, flags.patchesDir, flags.ignorePreflightErrors, flags.wait, flags.vLevel)
+		return KubeadmJoin(c, flags.usePhases, flags.copyCertsMode, flags.discoveryMode, flags.kubeadmConfigVersion, flags.patchesDir, flags.ignorePreflightErrors, flags.wait, flags.vLevel)
 	},
 	"kubeadm-upgrade": func(c *status.Cluster, flags *RunOptions) error {
 		return KubeadmUpgrade(c, flags.upgradeVersion, flags.patchesDir, flags.wait, flags.vLevel)
@@ -145,6 +145,13 @@ func IgnorePreflightErrors(ignorePreflightErrors string) Option {
 	}
 }
 
+// KubeadmConfigVersion option sets the kubeadm config version for the kubeadm commands
+func KubeadmConfigVersion(kubeadmConfigVersion string) Option {
+	return func(r *RunOptions) {
+		r.kubeadmConfigVersion = kubeadmConfigVersion
+	}
+}
+
 // RunOptions holds options supplied to actions.Run
 type RunOptions struct {
 	kubeDNS               bool
@@ -156,6 +163,7 @@ type RunOptions struct {
 	vLevel                int
 	patchesDir            string
 	ignorePreflightErrors string
+	kubeadmConfigVersion  string
 }
 
 // DiscoveryMode defines discovery mode supported by kubeadm join
--- a/kinder/pkg/cluster/manager/actions/kubeadm-config.go
+++ b/kinder/pkg/cluster/manager/actions/kubeadm-config.go
@@ -33,6 +33,7 @@ import (
 // kubeadmConfigOptionsall stores all the kinder flags that impact on the kubeadm config generation
 type kubeadmConfigOptions struct {
 	kubeDNS       bool
+	configVersion string
 	copyCertsMode CopyCertsMode
 	discoveryMode DiscoveryMode
 }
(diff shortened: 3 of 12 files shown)


kinder/pkg/kubeadm/config.go
```diff
@@ -27,17 +27,11 @@ import (
...
-// Config returns a kubeadm generated using the config API version corresponding
-// to the kubeadmVersion and with the customizable settings based on data
-func Config(kubeadmVersion *K8sVersion.Version, data ConfigData) (config string, err error) {
-	// gets the config version corresponding to a kubeadm version
-	kubeadmConfigVersion, err := getKubeadmConfigVersion(kubeadmVersion)
-	if err != nil {
-		return "", err
-	}
-
+// Config returns a kubeadm generated using the config API version
```
neolit123, line 30:
```suggestion
// Config returns a kubeadm config generated using the config API version
```

kinder/pkg/kubeadm/config.go
```diff
@@ -66,13 +60,15 @@ func Config(kubeadmVersion *K8sVersion.Version, data ConfigData) (config string,
...
 }
 
-// getKubeadmConfigVersion returns the kubeadm config version corresponding to a Kubernetes kubeadmVersion
-func getKubeadmConfigVersion(kubeadmVersion *K8sVersion.Version) (string, error) {
-	// returns the corresponding config version
-	// nb v1alpha1 (that is Kubernetes v1.10.0) is out of support,
-	// v1alpha2 (that is Kuberntes v1.11.0) is out of support.
-	// v1alpha3 (that is Kuberntes v1.13.0) is out of support.
+// GetKubeadmConfigVersion returns the kubeadm config version.
+// It returns the kubeadmConfigVersion specified by user first when that is available.
```
neolit123, line 64:
```suggestion
// It returns the kubeadmConfigVersion specified by the user first when that is available.
```

kinder/pkg/kubeadm/config.go
```diff
@@ -66,13 +60,15 @@ func Config(kubeadmVersion *K8sVersion.Version, data ConfigData) (config string,
...
+// Otherwise, returns the kubeadm config version corresponding to a Kubernetes kubeadmVersion.
+func GetKubeadmConfigVersion(kubeadmVersion *K8sVersion.Version, kubeadmConfigVersion string) (string, error) {
```
neolit123, line 66: why not remove `kubeadmVersion *K8sVersion.Version` from this function's arguments at this point + (also update the comment above the function)? it is not used.

kinder/pkg/kubeadm/kubednspatch.go
```diff
@@ -19,21 +19,13 @@ package kubeadm
 import (
```
neolit123, line 19: we forgot to remove kube-dns support from kinder.

here is an issue to track this:
https://github.com/kubernetes/kubeadm/issues/2484

neolit123:
thanks @SataQiu 
added some minor comments, mostly.

i think we can close this issue with the PR:
https://github.com/kubernetes/kubeadm/issues/2249
(`fixes ...`)

as part of this https://github.com/kubernetes/kubeadm/issues/2428 my proposal is to:
- drop v1beta1
- enable v1beta3 by default
- pass the flag with v1beta2 for all jobs that test kubeadm versions older than 1.22

kinder/pkg/kubeadm/kubednspatch.go
SataQiu: Yes, we can do this with another PR.

kinder/pkg/kubeadm/config.go
SataQiu: As this function has no complicated logic, I prefer to remove it completely.
We can just use a constant as the default kubeadm config version.

SataQiu:
> thanks @SataQiu
> added some minor comments, mostly.
> 
> i think we can close this issue with the PR:
> #2249
> (`fixes ...`)
> 
> as part of this #2428 my proposal is to:
> 
> * drop v1beta1
> * enable v1beta3 by default
> * pass the flag with v1beta2 for all jobs that test kubeadm versions older than 1.22

Good idea! Let's do this with the following PRs.

kinder/cmd/kinder/do/do.go
```diff
@@ -120,6 +121,12 @@ func NewCommand() *cobra.Command {
 		"ignore-preflight-errors", constants.KubeadmIgnorePreflightErrors,
 		"list of kubeadm preflight errors to skip",
 	)
+	cmd.Flags().StringVar(
+		&flags.KubeadmConfigVersion,
+		"kubeadm-config-version", constants.DefaultKubeadmConfigVersion,
+		fmt.Sprintf("the kubeadm config version to be used for init, join and upgrade; use one of %s",
```
fabriziopandini, line 127: What about adding: "if not set, kubeadm will automatically choose the Kubeadm config version according to the Kubernetes version in use"

kinder/pkg/cluster/manager/actions/actions.go
```diff
@@ -238,6 +246,25 @@ func ValidateCopyCertsMode(t CopyCertsMode) error {
 	return nil
 }
 
+// SupportedKubeadmConfigVersion returns the list of supported kubeadm config version
+func SupportedKubeadmConfigVersion() []string {
+	return []string{
+		"v1beta1",
+		"v1beta2",
```
fabriziopandini, line 253: should this include v1beta3 now? also, should we add a deprecation notice on v1beta1 - assuming we should keep this around for testing older versions -

kinder/pkg/cluster/manager/actions/actions.go
```diff
@@ -238,6 +246,25 @@ func ValidateCopyCertsMode(t CopyCertsMode) error {
...
+	}
+}
+
+// ValidateKubeadmConfigVersion validates a kubeadm config version
+func ValidateKubeadmConfigVersion(v string) error {
```
fabriziopandini, line 258: This is a little bit trickier than this, because the kubeadm config version is linked to the kubeadm version we are using. Not sure we can really catch this during flag validation, but let's make sure to keep this in my down the path

kinder/pkg/cluster/manager/actions/kubeadm-config.go
```diff
@@ -181,9 +183,10 @@ func getKubeadmConfig(c *status.Cluster, n *status.Node, data kubeadm.ConfigData
 	if err != nil {
 		return "", err
 	}
+	log.Debugf("kubeadm version %s", kubeadmVersion)
```
fabriziopandini, line 186: e.g This could be a good place to validate the kubeadm version actually supports the requested config version.
Also, we should make sure that if the config version is not specified, kinder uses a default config version like it is happening today

kinder/pkg/constants/constants.go
```diff
@@ -126,4 +126,7 @@ var (
 const (
 	// KinderVersion is the kinder CLI version
 	KinderVersion = "0.1.0-alpha.3"
+
+	// DefaultKubeadmConfigVersion indicates the default config API version that kubeadm uses
+	DefaultKubeadmConfigVersion = "v1beta2"
```
fabriziopandini, line 131: Uhmm, we can't have a DefaultKubeadmConfigVersion, because this depends by the kubeadm version we are using.
Eg. for kubeadm v1.22 it is v1beta3, for older version v1beta2 (and for older older versions v1beta1, but I'm not sure they are still in the support skew)

kinder/cmd/kinder/do/do.go
SataQiu: make sense!

kinder/pkg/cluster/manager/actions/actions.go
SataQiu: Emm... Maybe it's better to validate the config version through `Kubeadm` rather than `Kinder`?

kinder/pkg/kubeadm/config.go
```diff
@@ -66,14 +60,12 @@ func Config(kubeadmVersion *K8sVersion.Version, data ConfigData) (config string,
...
-func getKubeadmConfigVersion(kubeadmVersion *K8sVersion.Version) (string, error) {
-	// returns the corresponding config version
-	// nb v1alpha1 (that is Kubernetes v1.10.0) is out of support,
-	// v1alpha2 (that is Kuberntes v1.11.0) is out of support.
-	// v1alpha3 (that is Kuberntes v1.13.0) is out of support.
-
-	return "v1beta2", nil
+// GetKubeadmConfigVersion returns the kubeadm config version corresponding to a Kubernetes kubeadmVersion
+func GetKubeadmConfigVersion(kubeadmVersion *K8sVersion.Version) string {
+	if kubeadmVersion.AtLeast(K8sVersion.MustParseSemantic("v1.22.0")) {
```
SataQiu, line 65: @neolit123 @fabriziopandini 
I'm not sure what to do here, given the fact that `Kinder` doesn't currently support v1beta3.

neolit123: i think i agree that we have to add the v1beta3 support first before this PR.

i see you have:
https://github.com/kubernetes/kubeadm/pull/2488

xref https://github.com/kubernetes/kubeadm/issues/2428

kinder/pkg/cluster/manager/actions/actions.go
fabriziopandini: Kinder can "hardcode" this, it requires the same info you already used in `GetKubeadmConfigVersion` but in this case you have to check if the version provided by the user is supported by the Kubernetes version.

However, I'm fine also if we didn't validate consistence of config version vs kubeadm version  and just assume that a user playing with kinder knows which version to use.

fabriziopandini:
/lgtm
/hold
For Lubomir to take a look after latest changes

kinder/pkg/kubeadm/config.go
```diff
@@ -66,14 +60,15 @@ func Config(kubeadmVersion *K8sVersion.Version, data ConfigData) (config string,
...
-	// nb v1alpha1 (that is Kubernetes v1.10.0) is out of support,
-	// v1alpha2 (that is Kuberntes v1.11.0) is out of support.
-	// v1alpha3 (that is Kuberntes v1.13.0) is out of support.
-
-	return "v1beta3", nil
+// GetKubeadmConfigVersion returns the kubeadm config version corresponding to a Kubernetes kubeadmVersion
+func GetKubeadmConfigVersion(kubeadmVersion *K8sVersion.Version) string {
+	// v1alpha1 (that is Kubernetes v1.10.0) is out of support
+	// v1alpha2 (that is Kubernetes v1.11.0) is out of support
+	// v1alpha3 (that is Kubernetes v1.13.0) is out of support
```
neolit123, line 67: i think we can remove these comments.

kinder/pkg/kubeadm/config.go
```diff
@@ -66,14 +60,15 @@ func Config(kubeadmVersion *K8sVersion.Version, data ConfigData) (config string,
...
+	if kubeadmVersion.Major() > 1 || kubeadmVersion.Minor() >= 22 {
```
neolit123, line 68:
```suggestion
	if kubeadmVersion.Compare("v1.22.0") >= 0 {
```

you can do this, which is technically the accurate check.
but the accuracy does not matter much since we have only MAJOR == 1 for k8s.
https://github.com/kubernetes/apimachinery/blob/master/pkg/util/version/version.go#L319

neolit123:
looks like we have a problem where a 1.20 e2e test is using v1beta3:
https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-e2e-kubeadm-kinder-kubelet-1-19-on-1-20/1397633402328322048/artifacts/task-03-init-log.txt
https://testgrid.k8s.io/sig-cluster-lifecycle-kubeadm#kubeadm-kinder-kubelet-1-19-on-1-20

neolit123:
ok, looks like we broke all kubeadm tests for old versions, since v1beta3 is not available on < 1.22:
https://github.com/kubernetes/kubeadm/pull/2488

the change in this PR is needed, so i'm removing the hold.

/hold cancel

PR #2495, opened 2021-05-31
kinder: pin the CR cgroup driver to "systemd"

neolit123:
Pin the driver to "systemd" in:
- the Docker / containerd config files during image alter
- KubeletConfiguration that kubeadm deploys

xref https://github.com/kubernetes/kubeadm/issues/874

PR #2497, opened 2021-06-01
[WIP] kinder: pass --kubeadm-config-version flag with v1beta2 for all the jobs that test kubeadm/kubernetes versions older than 1.22

SataQiu:
If we are testing the old version of `Kubernetes` with the latest `Kubeadm`, we should use the old compatible Kubeadm config. Otherwise, we may run into problems. 

FYI: https://testgrid.k8s.io/sig-cluster-lifecycle-kubeadm#kubeadm-kinder-latest-on-1-21

Ref: #2428

### DIFF
--- a/kinder/ci/tools/update-workflows/pkg/types.go
+++ b/kinder/ci/tools/update-workflows/pkg/types.go
@@ -68,10 +68,11 @@ type job struct {
 }
 
 type templateVars struct {
-	KubernetesVersion string
-	KubeletVersion    string
-	KubeadmVersion    string
-	InitVersion       string
+	KubernetesVersion    string
+	KubeletVersion       string
+	KubeadmVersion       string
+	KubeadmConfigVersion string
+	InitVersion          string
 
 	TargetFile   string
 	SkipVersions string
--- a/kinder/ci/tools/update-workflows/pkg/workflows.go
+++ b/kinder/ci/tools/update-workflows/pkg/workflows.go
@@ -27,6 +27,8 @@ import (
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	versionutil "k8s.io/apimachinery/pkg/util/version"
+
+	"k8s.io/kubeadm/kinder/pkg/kubeadm"
 )
 
 func processWorkflows(settings *Settings, cfg *jobGroup, oldestVer, minVer *versionutil.Version) error {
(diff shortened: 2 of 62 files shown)


neolit123:
there are two problems in https://testgrid.k8s.io/sig-cluster-lifecycle-kubeadm#kubeadm-kinder-latest-on-1-21

**1. the kubelet is failing due to cgroup drivers:**

https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-e2e-kubeadm-kinder-latest-on-1-21/1399542017624641536/artifacts/kinder-xony-control-plane-1/kubelet.log

> Jun 01 01:46:51 kinder-xony-control-plane-1 kubelet[1156]: E0601 01:46:51.112981    1156 server.go:292] "Failed to run kubelet" err="failed to run Kubelet: misconfiguration: kubelet cgroup driver: \"systemd\" is different from docker cgroup driver: \"cgroupfs\""

this should be fixed with:
https://github.com/kubernetes/kubeadm/pull/2495

**2. we are e2e testing for ClusterStatus:**

should be fixed by:
https://github.com/kubernetes/kubernetes/pull/102358

i think we can avoid using the --kubeadm-config-version flag for the time being as long as we have v1beta2 for versions that are older than 1.22. this is already being done in kinder and the test is passing in that regard.

fabriziopandini:
Uhmm, I'm not sure I understand 

> If we are testing the old version of Kubernetes with the latest Kubeadm, we should use the old compatible Kubeadm config. Otherwise, we may run into problems.

As far as I remember, kinder will use the right version of kubeadm config for the kubeadm/kubernetes version we are using, and this flag should be used only if we want to deviate from this default behavior and force tests to use different kubeadm config versions

SataQiu:
Thanks for your feedback @neolit123 @fabriziopandini 

As @neolit123  said, after kubernetes/kubernetes#102358 is merged, the problem I'm worried about should be fixed automatically.  We'd better solve the problem at the least cost.
/close

SataQiu:
> Uhmm, I'm not sure I understand
> 
> > If we are testing the old version of Kubernetes with the latest Kubeadm, we should use the old compatible Kubeadm config. Otherwise, we may run into problems.
> 
> As far as I remember, kinder will use the right version of kubeadm config for the kubeadm/kubernetes version we are using, and this flag should be used only if we want to deviate from this default behavior and force tests to use different kubeadm config versions

@fabriziopandini 
This is certainly no problem if we always keep the same version of Kubeadm and Kubernetes.
But we might use the higher version of Kubeadm to create a lower version of Kubernetes cluster.
In this scenario, we would choose the higher version of Kubeadm Config Version, but it may not pass the test of the lower version of Kubernetes e2e test.
From the log of https://testgrid.k8s.io/sig-cluster-lifecycle-kubeadm#kubeadm-kinder-latest-on-1-21, the Kubeadm version we used is 1.22.0-alpha.2.301+7e2b5f3325a2f7(Kubeadm Config Version v1beta3), but the Kubernetes version is v1.21.2-rc.0.24+d9aa7797177196.

PR #2517, opened 2021-06-25
Use /bin/bash instead of /bin/sh for verify-rootless task.

vinayakankugoyal:
tertgrid fails with the following error:

```
/bin/sh: 1: Bad substitution
chmod: cannot access '/tmp/verify-rootless.sh': No such file or directory
 exit status 1
```

https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-kubeadm#kubeadm-kinder-rootless-latest


neolit123 (commented):
/lgtm
/approve
/area kinder testing

PR #2520, opened 2021-06-29
kinder: use the "auto" mode for the "rootless" workflow

neolit123:
The "rootless" e2e test requires strict permissions
for some files like ca.crt. Apparently our existing
"manual" copy-certs method stomps permissions and
ownership ("docker cp" artifact?)

Use the "auto" mode which uses the native logic in kubeadm
that uses a Secret as the transport medium of certs/keys
between control plane nodes. It also stomps permissions, but
has some logic to write private keys with 600 and public
keys/certs with 644, which is sufficient for the "rootless" use case.


neolit123:
self LGTM-ing as the k/k PR merged and this is needed for a green e2e run:
https://github.com/kubernetes/kubernetes/pull/103313

PR #2521, opened 2021-06-30
Use pgrep -f to do a full match and use regular expression to make su…

vinayakankugoyal:
…re that the command start with the desired name.


neolit123 (commented):
/lgtm
/approve

PR #2528, opened 2021-07-17
use k8s-release-dev instead of kubernetes-release-dev

spiffxp:
this was a straight search-replace rather than attempting to converge
toward dl.k8s.io because way too much of this is stale. I tried editing
to update stale info at first, but someone who knows the state of things
needs to do this.

e.g. many of the uris in here are already dead, let alone things like
/bazel and /ci-cross not existing in gs://k8s-release-dev

This is part of https://github.com/kubernetes/k8s.io/issues/2318


neolit123 (commented):
thanks,

logged this issue for tracking:
https://github.com/kubernetes/kubeadm/issues/new

/lgtm
/approve

spiffxp:
@neolit123 just double-checking, was there a reason you removed the `approved` label?

neolit123:
> @neolit123 just double-checking, was there a reason you removed the approved label?

no, this has happened for me before, so either there is a github bug, or i'm unchecking the label (i don't think so).

Issue #2547, opened 2021-08-16
patches: patch support e2e are still broken


neolit123:
we made the e2e tests use the v1beta3.Init|JoinConfiguration.patches.directory feature:
https://github.com/kubernetes/kubeadm/pull/2545
https://github.com/kubernetes/kubeadm/pull/2537

but the CI is failing:

https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-kubeadm#kubeadm-kinder-patches-latest
https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-e2e-kubeadm-kinder-patches-latest/1427291038157901824/build-log.txt

looks like the patches are being copied to the nodes:
> time="15:35:27" level=debug msg="Running: [docker cp /tmp/kubeadm-patches/kube-scheduler0+strategic.yaml kinder-patches-control-plane-1:/tmp/kubeadm-patches/kube-scheduler0+strategic.yaml]"
Error: No such container:path: kinder-patches-control-plane-1:/tmp/kubeadm-patches

the kubeadm config file contains the patches.directory path:
> time="15:36:47" level=debug msg="generated config:\napiVersion: kubeadm.k8s.io/v1beta3\ncontrolPlane:\n  localAPIEndpoint:\n    advertiseAddress: 172.17.0.4\n    bindPort: 6443\ndiscovery:\n  bootstrapToken:\n    apiServerEndpoint: 172.17.0.5:6443\n    token: abcdef.0123456789abcdef\n    unsafeSkipCAVerification: true\nkind: JoinConfiguration\nmetadata:\n  name: config\nnodeRegistration:\n  criSocket: /var/run/dockershim.sock\n  kubeletExtraArgs:\n    node-ip: 172.17.0.4\npatches:\n  directory: /tmp/kubeadm-patches\n"
time="15:36:47" level=debug msg="Running: [docker cp /tmp/kinder-patches-control-plane-2-283719042 kinder-patches-control-plane-2:/kind/kubeadm.conf]"
time="15:36:47" level=debug msg="Running: [docker exec kinder-patches-control-plane-2 kubeadm version -o=short]"

but then kubeadm init finds no patches in the given directory /tmp/kubeadm-patches:
> [patches] Reading patches from path "/tmp/kubeadm-patches"
> ...no patches were found here.

so either the patches were not copied correctly, or there is a bug in kubeadm.
will try to investigate locally.

neolit123:
cc @pacoxu

[...]

PR #2548, opened 2021-08-16
kinder: fix failing patches e2e tests

neolit123:
- Looks like /tmp is not a good location to copy files to.
    Either they silently fail to copy to /tmp or they throw an
    error that a /tmp/sub-dir like "kinder-patches" is not a directory
    (but it is). Searching for the related errors online:
      - Error response from daemon: extraction point is not a directory
      - Error: No such container:path: kinder-patches-control-plane-1:/tmp/kubeadm-patches
    does not provide much context on why this is happening.
    Use "/kinder/patches" instead of "/tmp...".
    
    - Include missing error return in copyPatchesToNode()
    
    - Remove host OS check about the patch directory. We don't need this check
    because an earlier step should copy the patches from host to container.
    If the patch directory in the container is different from the patch
    directory on the host it would fail.
    
    - Given patches.directory is always part of the kubeadm config,
    always create the patch directory with copyPatchesToNode(). Skip
    copying patches if the source directory is not defined.

fixes: https://github.com/kubernetes/kubeadm/issues/2547

### DIFF
--- a/kinder/pkg/cluster/manager/actions/kubeadm-init.go
+++ b/kinder/pkg/cluster/manager/actions/kubeadm-init.go
@@ -40,11 +40,8 @@ import (
 func KubeadmInit(c *status.Cluster, usePhases bool, copyCertsMode CopyCertsMode, kubeadmConfigVersion, patchesDir, ignorePreflightErrors, featureGates string, wait time.Duration, vLevel int) (err error) {
 	cp1 := c.BootstrapControlPlane()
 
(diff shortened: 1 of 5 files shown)


neolit123:
cc @pacoxu you can check what changes i did here.

neolit123:
self LGTM-ing to unblock e2e. these tests have been failing for a while.
https://github.com/kubernetes/kubeadm/issues/2547

PR #2555, opened 2021-08-19
github/workflows: add action for a "/support" command

neolit123:
Users tend to skip our notes about not using the issue tracker
for support questions.

This change adds a GitHub action handler for a `/support`
command.

Commenting with the command on an issue will:
- Provide a friendly message where to find support links.
- Label the issue with `kind/support`.
- Close the issue.

### DIFF
--- a/.github/workflows/support-command.yaml
+++ b/.github/workflows/support-command.yaml
@@ -1,0 +1,34 @@
+on: issue_comment
+name: handle the /support command
+jobs:
+  support_comment:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


neolit123:
the message looks like this:

```
Hello, @neolit123 :robot: :wave:

You seem to have troubles using Kubernetes and kubeadm.
Note that our issue trackers **should not** be used for providing support to users.
There are special channels for that purpose.

Please see:
- https://github.com/kubernetes/kubeadm#support
```

pacoxu (approved):
/lgtm
Nice to have 😄

Issue #2551, opened 2021-08-18
add import order tool to this repository


neolit123:
xref https://github.com/kubernetes/kubeadm/issues/2515

the tool for import ordering was rejected from k/k, but we can at least archive it in this repository.

a good location for the tool / scripts is /kinder/hack:
https://github.com/kubernetes/kubernetes/pull/103981

which will effectively also manage the imports of kinder.

cc @xinydev as you have the source code for it.

fabriziopandini:
Q: is there a reason why are we adding more rules than Kubernetes/than the ones that are enforced by available linterts? I other words, what I'm trying to understand what is what is is the advantage in becoming opinionated about imports for this repository only and if this is worth maintaining a linter in this repository.

PR #2557, opened 2021-08-23
Archive imports order tool to this repository

xinydev:
Fixes: https://github.com/kubernetes/kubeadm/issues/2551 

update the local path from  `k8s.io/kubernetes ` to `k8s.io/kubeadm`


kinder/hack/update-imports-order.sh
```diff
@@ -0,0 +1,50 @@
...
+
+# shellcheck source=/dev/null
+source "$(dirname "$0")/utils.sh"
+# cd to the root path
+cd_root_path
+
+go install k8s.io/kubeadm/kinder/hack/orderimports
+
+
+cd "${KUBE_ROOT}"
```
pacoxu, line 31: https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/hack/update-bazel.sh#L20
```
KUBE_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
```
I got an error when testing this PR.
```
➜  kinder git:(order-imports) hack/update-imports-order.sh
hack/update-imports-order.sh:行31: KUBE_ROOT：未绑定的变量
```

xinydev: Thanks a lot. I forgot to delete this line of code. Fixed

kinder/hack/orderimports/OWNERS
```diff
@@ -0,0 +1,10 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+reviewers:
+  - dims
+  - neolit123
+  - xinydev
+approvers:
+  - dims
+  - neolit123
+  - xinydev
```
neolit123, line 10: noting that having this subfolder owners file seems fine even if this is now in the kubeadm repo.

kinder/hack/verify-imports-order.sh
```diff
@@ -0,0 +1,60 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# shellcheck source=/dev/null
+source "$(dirname "$0")/utils.sh"
+# cd to the root path
+cd_root_path
+
+go install k8s.io/kubeadm/kinder/hack/orderimports
```
neolit123, line 30: can we avoid `go install`?
i don't think any of the scripts in kinder/hack use it.

kinder/hack/update-imports-order.sh
```diff
@@ -0,0 +1,47 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# shellcheck source=/dev/null
+source "$(dirname "$0")/utils.sh"
+# cd to the root path
+cd_root_path
+
+go install k8s.io/kubeadm/kinder/hack/orderimports
```
neolit123, line 28: same here.

kinder/hack/orderimports/orderimports.go
```diff
@@ -0,0 +1,463 @@
...
+	for _, spec := range d.Specs {
+		imp := spec.(*ast.ImportSpec)
+		importPath := strings.Replace(imp.Path.Value, "\"", "", -1)
+		parts := strings.Split(importPath, "/")
+		if !strings.Contains(parts[0], ".") {
+			// standard library
+			stdlibImports = append(stdlibImports, imp)
+		} else if strings.HasPrefix(importPath, "k8s.io/kubeadm") {
+			// local imports
+			localImports = append(localImports, imp)
```
neolit123, line 192: i think we should make the tool support any local prefix, potentially via a new flag.
this means that i could build the tool and call:
`orderimports.go -p k8s.io/kubeadm ...`

which would order with `k8s.io/kubeadm` being the local import path prefix.
autodetecting this seems trickier, but perhaps you have ideas @xinydev ?

(not sure if `-p` is right package short flag  vs e.g. `-local`)

in theory this would allow us to use the tool for any repo under k8s.

xinydev: adding a new flag is a good idea!

neolit123:
The CI is failing due to the go 1.17 update. Should be resolved later today
/ tomorrow.

neolit123:
tests should pass now

/retest

PR #2582, opened 2021-10-02
Update ha-considerations.md

MagonBC:
When APISERVER_DEST_PORT is different from 6443, we should use the appropriate port with each one of the two CURL commands. If haproxy is set to bind on a specific VIP (for example: bind 192.168.1.4:8443), the first curl will fail and the failover will take place even if the VIP is not holded by the current machine.

### DIFF
--- a/docs/ha-considerations.md
+++ b/docs/ha-considerations.md
@@ -91,7 +91,7 @@ errorExit() {
     exit 1
 }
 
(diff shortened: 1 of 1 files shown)


MagonBC:
Double checked. current config is ok. No need to specify the APISERVER_SRC_PORT if the frontend apiserver has *:8443 bind (haproxy.cfg). I will close the request.

PR #2593, opened 2021-10-29
Replace iouitl with io and os

ahrtr:
The ioutil package has already been deprecated in golang 1.16, please see [go1.16#ioutil](https://golang.org/doc/go1.16#ioutil). So we also need to replace all the ioutil with os and io.


ahrtr:
> /ok-to-test changes seems ok, but test failures should be investigated

@fabriziopandini the test failure was fixed.

pacoxu:
/lgtm
test passed now
/assign @neolit123

neolit123 (commented):
Thanks
/approve

PR #2606, opened 2021-11-18
Update ha-considerations.md

thebsdbox:
Big-rewrite of the Kube-vip.io section. It hasn't been updated in ~1 year.

### DIFF
--- a/docs/ha-considerations.md
+++ b/docs/ha-considerations.md
@@ -255,60 +255,64 @@ With the services up, now the Kubernetes cluster can be bootstrapped using `kube
 
 ## kube-vip
 
-As an alternative to the more "traditional" approach of `keepalived` and `haproxy`, [kube-vip](https://kube-vip.io/) implements both management of a virtual IP and load balancing in one service. Similar to option 2 above, `kube-vip` will be run as a static pod on the control plane nodes.
+As an alternative to the more "traditional" approach of `keepalived` and `haproxy`, [kube-vip](https://kube-vip.io/) implements both management of a virtual IP and load balancing in one service. It can be implemented in bother layer2 (with ARP, and `leaderElection`) or layer3 utilising BGP peering. Similar to option 2 above, `kube-vip` will be run as a static pod on the control plane nodes.
 
 Like with `keepalived`, the hosts negotiating a virtual IP need to be in the same IP subnet. Similarly, like with `haproxy`, stream-based load-balancing allows TLS termination to be handled by the API Server instances behind it.
 
-The configuration file `/etc/kube-vip/config.yaml` looks like this:
-```yaml
-localPeer:
-  id: ${ID}
-  address: ${IPADDR}
-  port: 10000
-remotePeers:
-- id: ${PEER1_ID}
-  address: ${PEER1_IPADDR}
-  port: 10000
-# [...]
-vip: ${APISERVER_VIP}
-gratuitousARP: true
-singleNode: false
-startAsLeader: ${IS_LEADER}
-interface: ${INTERFACE}
-loadBalancers:
-- name: API Server Load Balancer
-  type: tcp
-  port: ${APISERVER_DEST_PORT}
-  bindToVip: false
-  backends:
-  - port: ${APISERVER_SRC_PORT}
-    address: ${HOST1_ADDRESS}
-  # [...]
+**NOTE**
+`kube-vip` requires access to the API server, especially during a cluster initialisation (during the `kubeadm init` phase). At this point the `admin.conf` is the only kubeconfig that is available to `kube-vip` to authenticate and communicate with the API-server. Post cluster stand up it is recommended that a user sign a custom client kubeconfig and rotate it manually on expiration.
+
+### Generating a Manifest
+
+This section details creating a number of manifests for various use cases
+
+#### Set configuration details
+
+```
+export VIP=192.168.0.40`
+export INTERFACE=<interface>
 ```
 
-The `bash` style placeholders to expand are these:
-- `${ID}` the current host's symbolic name
-- `${IPADDR}` the current host's IP address
-- `${PEER1_ID}` a symbolic name for the first vIP peer
-- `${PEER1_IPADDR}` IP address for the first vIP peer
-- entries (`id`, `address`, `port`) for additional vIP peers can follow
-- `${APISERVER_VIP}` is the virtual IP address negotiated between the `kube-vip` cluster hosts.
-- `${IS_LEADER}` is `true` for exactly one leader and `false` for the rest
-- `${INTERFACE}` is the network interface taking part in the negotiation of the virtual IP, e.g. `eth0`.
-- `${APISERVER_DEST_PORT}` the port through which Kubernetes will talk to the API Server.
-- `${APISERVER_SRC_PORT}` the port used by the API Server instances
-- `${HOST1_ADDRESS}` the first load-balanced API Server host's IP address
-- entries (`port`, `address`) for additional load-balanced API Server hosts can follow
+### Configure to use a container runtime
 
-To have the service started with the cluster, now the manifest `kube-vip.yaml` needs to be placed in `/etc/kubernetes/manifests` (create the directory first). It can be generated using the `kube-vip` docker image:
+#### Get latest version
+
+ We can parse the GitHub API to find the latest version (or we can set this manually)
+
+`KVVERSION=$(curl -sL https://api.github.com/repos/kube-vip/kube-vip/releases | jq -r ".[0].name")`
+
+or manually:
+
+`export KVVERSION=vx.x.x`
+
+The easiest method to generate a manifest is using the container itself, below will create an alias for different container runtimes.
+
+#### containerd
+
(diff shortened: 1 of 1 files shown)


docs/ha-considerations.md
```diff
@@ -255,60 +255,140 @@ With the services up, now the Kubernetes cluster can be bootstrapped using `kube
...
-  # [...]
+### Generating a Manifest
+
+This section details creating a number of manifests for various use cases
+
+#### Set configuration details
+
+`export VIP=192.168.0.40`
+
+`export INTERFACE=<interface>`
```
pacoxu, line 270:
````suggestion
```
export VIP=192.168.0.40`

export INTERFACE=<interface>`
```
````

File: docs/ha-considerations.md [hunk 2]
```diff
@@ -255,60 +255,140 @@ With the services up, now the Kubernetes cluster can be bootstrapped using `kube
...
+      path: /etc/kubernetes/admin.conf
+    name: kubeconfig
+status: {}
+```
+
+### BGP
+
+This configuration will create a manifest that will start `kube-vip` providing **controlplane** and **services** management. **Unlike** ARP, all nodes in the BGP configuration will advertise virtual IP addresses. 
+
+**Note** we bind the address to `lo` as we don't want multiple devices that have the same address on public interfaces. We can specify all the peers in a comma seperate list in the format of `address:AS:password:multihop`.
```
pacoxu, line 374: nit: s/seperate/seperated/w

docs/ha-considerations.md
```diff
@@ -317,27 +397,59 @@ metadata:
...
+    - name: bgp_peeraddress
+    - name: bgp_peerpass
+    - name: bgp_peeras
+      value: "65000"
+    - name: bgp_peers
+      value: 192.168.0.10:65000::false,192.168.0.11:65000::false
+    - name: vip_address
+      value: 192.168.0.40
+    image: ghcr.io/kube-vip/kube-vip:v0.4.0
+    imagePullPolicy: Always
```
pacoxu, line 432: The imagePullPolicy is `IfNotPresent` by default. Should we change the default value?

docs/ha-considerations.md
```diff
@@ -317,27 +397,59 @@ metadata:
...
+  - hostnames:
+    - kubernetes
+    ip: 127.0.0.1
   hostNetwork: true
   volumes:
   - hostPath:
-      path: /etc/kube-vip/config.yaml
-    name: config
+      path: /etc/kubernetes/admin.conf
+    name: kubeconfig
```
pacoxu, line 452: For Principle of Least Privilege, we'd better not use admin config.

pacoxu:
Only some nit from my side.

docs/ha-considerations.md
thebsdbox: This is a requirement, as this static pod is brought up as the control plane is started and needs to be able to speak with the API server as it starts.

docs/ha-considerations.md
thebsdbox: hmm, this is in the code of Kube-vip as it generates the manifests.. so the documentation is correct, but the default behaviour of Kube-vip manifest generation will need to change.

docs/ha-considerations.md
pacoxu: LGTM except this one. 
/cc @neolit123

thebsdbox: No worries: the `/etc/kubernetes/admin.conf` is passed so that the go-client can speak to the API server during the `kubeadm init` and perform a leaderElection and bring up the HA virtual IP address.

neolit123:
Thanks for these changes. I was away for a week and will review as soon as
I can.

docs/ha-considerations.md
neolit123: mounting the admin.conf in the container grants it super power over the cluster.
also the admin.conf certificates cannot be revoked currently and cannot be gated by RBAC.

i'd say at least ghcr.io/kube-vip/kube-vip should be a distroless image to avoid distro security leaks in the image. is it distroless already? this will make it match what we have for the other component images like KCM.

in an ideal world the client that kube-vip has should be limited to only the actions it needs - control the LE for the apiserver. signing a new kubeconfig should be a few additional commands.

```
kubeadm kubeconfig user --client-name=kube-vip --org kube-vip ...
kubectl ... # add RBAC rules for kube-vip group
...
```

but i can see that this will complicate the UX around kube-vip because it means users would also have to rotate the new kubeconfig and `kubeadm certs renew` will not do it for them

neolit123 (commented):
i like the update and the automation that this adds for manifest generation 
@thebsdbox

we need to figure out what to do about the admin.conf mount and potentially moving the kube-vip image to distroless.

docs/ha-considerations.md
neolit123: i can see us having this with admin.conf by default but adding a NOTE that if users want to limit the access of kube-vip they should sign a custom client kubeconfig and rotate it manually on expiration.

thebsdbox: kube-vip is not distroless it uses `SCRATCH`.

docs/ha-considerations.md
```diff
@@ -255,60 +255,59 @@ With the services up, now the Kubernetes cluster can be bootstrapped using `kube
...
+`KVVERSION=$(curl -sL https://api.github.com/repos/kube-vip/kube-vip/releases | jq -r ".[0].name")`
+
+or manually:
+
+`export KVVERSION=vx.x.x`
+
+The easiest method to generate a manifest is using the container itself, below will create an alias for different container runtimes.
+
+#### containerd
+`alias kube-vip="ctr run --rm --net-host ghcr.io/kube-vip/kube-vip:$KVVERSION vip /kube-vip"`
```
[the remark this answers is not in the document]
thebsdbox: The problem is that we need to speak with the API-server during cluster standup  ¯\_(ツ)_/¯.. as mentioned kube-vip uses `SCRATCH`.

docs/ha-considerations.md
```diff
@@ -255,60 +255,64 @@ With the services up, now the Kubernetes cluster can be bootstrapped using `kube
...
-- name: API Server Load Balancer
-  type: tcp
-  port: ${APISERVER_DEST_PORT}
-  bindToVip: false
-  backends:
-  - port: ${APISERVER_SRC_PORT}
-    address: ${HOST1_ADDRESS}
-  # [...]
+**NOTE**
+`kube-vip` requires access to the API server, especially during a cluster initialisation (during the `kubeadm init` phase). At this point the `admin.conf` is the only kubeconfig that is available to `kube-vip` to authenticate and communicate with the API-server. Post cluster stand up it is recommended that a user sign a custom client kubeconfig and rotate it manually on expiration.
```
neolit123, line 263: this should suffice, thanks

neolit123: >  is passed so that the go-client can speak to the API server during the kubeadm init and perform a leaderElection and bring up the HA virtual IP address.

what RBAC is needed by kube-vip exactly - i.e. what API calls does it make?
based on that we can include a minimal ClusterRole(Binding):

```
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: kube-vip
rules:
- apiGroups:
  - coordination.k8s.io
  resources:
  - leases
  verbs:
  - get
  - list
  - watch
  - create
  - update
  - patch
  - delete
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: kube-vip
subjects:
- kind: User
  name: kube-vip
  apiGroup: rbac.authorization.k8s.io
roleRef:
  kind: ClusterRole
  name: kube-vip
  apiGroup: rbac.authorization.k8s.io
```

and mention this example command:
```
kubeadm kubeconfig user --config=kubeadminit.conf --client-name=kube-vip
```

/approve
/tide merge-method-squash

neolit123: > /tide merge-method-squash

doesn't work for some reason, so please keep the commits squashed to 1.

thebsdbox:
Squashed ! :-)

neolit123:
Thanks. Did you see this question?

> what RBAC is needed by kube-vip exactly - i.e. what API calls does it make?

https://github.com/kubernetes/kubeadm/pull/2606#discussion_r767061810

docs/ha-considerations.md
thebsdbox: There are a few (mainly all required for ARP or services):

https://kube-vip.io/manifests/rbac.yaml
largely contains all of the requirements when it runs as a daemonset

thebsdbox: Although without the `admin.conf` being bound mounted into the static pod how will a client-go client auth/interact with the API server?

neolit123: It still needs client credentials but if i understand the problem correctly the permissions need to be pre baked. I.e. rbac rules need to exist at api server start time.

Using admin.conf entirely bypasses rbac because it signed for the system:masters group.

neolit123:
/lgtm
Thanks

PR #2643, opened 2022-01-27
kinder: wait for node containers to start

neolit123:
there are still some flakes in the kinder jobs:
https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-kubeadm#kubeadm-kinder-1-21

```
time="15:56:56" level=debug msg="Writing node settings at /kinder/node-settings.yaml (attempt 22)..."
time="15:56:56" level=debug msg="Running: [docker exec kinder-regular-control-plane-1 mkdir -p /kinder]"
...
Error: failed to create cluster: failed to create directory /kinder: exit status 1
time="15:57:00" level=error msg="failed to create directory /kinder: exit status 1"
 exit status 1
```

Kinder node containers are started in parallel.
In some cases trying to write settings to the nodes
right after that (via e.g. mkdir -p /kinder ...) can fail,
without a definitive reason. Also this is not reproducible
locally it seems, so perhaps it happens in CI (slower).
A suspicion here is that the parallel start can sometime
render containers as never having a Running=true state.

- Add serial wait functions that wait for the kinder node
containers to have Running=true.
- Don't write both the node/cluster settings:
  - node settings are empty.
  - cluster settings have IPfamily but this was already disabled before.

If the flakes continue and some nodes have Running=false,
one further option can be to try creating the containers serially.

### DIFF
--- a/kinder/pkg/cluster/manager/create.go
+++ b/kinder/pkg/cluster/manager/create.go
@@ -19,9 +19,12 @@ package manager
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
+
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	"k8s.io/kubeadm/kinder/pkg/cluster/status"
 	"k8s.io/kubeadm/kinder/pkg/constants"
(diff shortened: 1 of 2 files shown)


kinder/pkg/cluster/manager/create.go
```diff
@@ -222,26 +225,56 @@ func createNodes(clusterName string, flags *CreateOptions) error {
...
 		return err
 	}
 
-	// writes to the nodes the cluster settings that will be re-used by kinder during the cluster lifecycle.
 	c.Settings = &status.ClusterSettings{
-		IPFamily: status.IPv4Family, // support for ipv6 is still WIP
-	}
-	if err := c.WriteSettings(); err != nil {
-		return err
+		IPFamily: status.IPv4Family, // only IPv4 is tested with kinder
```
neolit123, line 261: the diff is weird here, but basically this blob is preserved, just the commend is different

```
c.Settings = &status.ClusterSettings{
	IPFamily: status.IPv4Family, // only IPv4 is tested with kinder
}
```

kinder/pkg/cluster/manager/create.go
```diff
@@ -222,26 +225,56 @@ func createNodes(clusterName string, flags *CreateOptions) error {
...
+	//
+	// if err := c.WriteSettings(); err != nil {
+	// 	return err
+	// }
+	//
+	// for _, n := range c.K8sNodes() {
+	// 	if err := n.WriteNodeSettings(&status.NodeSettings{}); err != nil {
+	// 		return err
+	// 	}
+	// }
```
neolit123, line 277: comment out the settings writes on the side of the callers instead of inside the Write* functions.

kinder/pkg/cluster/status/node.go
```diff
@@ -233,30 +233,24 @@ func (n *Node) EtcdImage() (string, error) {
 	return n.etcdImage, nil
 }
 
-//const clusterSettingsPath = "/kinder/cluster-settings.yaml"
+const clusterSettingsPath = "/kinder/cluster-settings.yaml"
 
 // WriteClusterSettings stores in the node a set of cluster-wide settings that will be re-used
 // by kinder during the cluster lifecycle (after create)
 func (n *Node) WriteClusterSettings(settings *ClusterSettings) error {
```
neolit123, line 240: this function is being uncommented, but it's not used.

kinder/pkg/cluster/status/node.go
```diff
@@ -296,30 +289,18 @@ func (n *Node) WriteNodeSettings(settings *NodeSettings) error {
 		return errors.Wrapf(err, "failed to encode %s", nodeSettingsPath)
 	}
 
-	var lastError error
 	dir := filepath.Dir(nodeSettingsPath)
-	attempts := 0
```
neolit123, line 301: the poll here on directory creation / file write is being removed from this function because polling is now being done after node creation.

neolit123:
will watch the signal after this change...

Issue #1582, opened 2019-05-30
kubelet-config-x.y ConfigMap logic improvements


neolit123:
tracking issue in k/e + KEP: https://github.com/kubernetes/enhancements/issues/2915

1.23 Alpha:
- [x] deprecate the old format and introduce the new one controlled by UnversionedKubeletConfigMap:
https://github.com/kubernetes/kubernetes/pull/105741
- [x] add e2e tests in k/kubeadm:
https://github.com/kubernetes/kubeadm/pull/2601

1.24 Beta:
- [x] update KEP:
https://github.com/kubernetes/enhancements/pull/3120
- [x] flip the FG to enabled and graduate it to Beta, update the appropriate Beta TODOs introduced in the [original PR](https://github.com/kubernetes/kubernetes/pull/105741):
https://github.com/kubernetes/kubernetes/pull/108027
- [x] other k/k fixes:
https://github.com/kubernetes/kubernetes/pull/108336
https://github.com/kubernetes/kubernetes/pull/108128
https://github.com/kubernetes/kubernetes/pull/108127
- [x] cleanup k/kubeadm / update e2e tests:
https://github.com/kubernetes/kubeadm/pull/2654
- [x] docs: cleanup any kubelet-config-x.yy instances in at k/website
https://github.com/kubernetes/website/pull/31687

1.25 GA:
- [x] update KEP to GA:
https://github.com/kubernetes/enhancements/pull/3316
- [x] remove logic related to the feature gate and lock it to true. update the appropriate GA TODOs introduced in the [original PR](https://github.com/kubernetes/kubernetes/pull/105741):
https://github.com/kubernetes/kubernetes/pull/110327
https://github.com/kubernetes/kubernetes/pull/111146
- [x] remove e2e tests in k/kubeadm as the default e2e tests will exercise this
https://github.com/kubernetes/test-infra/pull/26461
https://github.com/kubernetes/kubeadm/pull/2702
- [x] k/website docs (update kubeadm-init page FG section)
https://github.com/kubernetes/website/pull/34159

1.26 post-GA:
- [x] remove the feature gate
https://github.com/kubernetes/kubernetes/pull/113448

-------------
EDIT: original issue description below

when a control-plane node is created by `init` using `kubernetesVersion: X.Y`, kubeadm will create a "versioned" ConfigMap called `kubelet-config-X.Y`.

[...]

PR #2654, opened 2022-02-09
kinder: add handling of the new/old kubelet-config map

neolit123:
With UnversionedKubeletConfigMap going Beta in 1.24
and enabled by default, we want to match what kubeadm is doing:
- look of the new name of CM
- if that fails fallback to the old CM name

leave a TODO to remove this extra handling once 1.24
goes out of support. This would mean that all tested
versions of kubeadm only have UnversionedKubeletConfigMap=true
and the new CM name.

xref https://github.com/kubernetes/kubeadm/issues/1582
xref https://github.com/kubernetes/kubernetes/pull/108027
(does not depend strictly on the this PR)

NOTE: i'm not touching the only e2e test workflow for this because it already excercises the case for upgrading from a state with old names -> new names.
https://github.com/kubernetes/kubeadm/blob/main/kinder/ci/workflows/unversioned-kubelet-cm-tasks.yaml#L113-L125
we can technically add a new workflow with the FG turned off and tested during init, but i don't think that is necessary.

### DIFF
--- a/kinder/pkg/cluster/manager/actions/waiter.go
+++ b/kinder/pkg/cluster/manager/actions/waiter.go
@@ -329,14 +329,30 @@ func staticPodHasVersion(pod, version string) func(c *status.Cluster, n *status.
 func kubeletHasRBAC(major, minor uint) func(c *status.Cluster, n *status.Node) bool {
 	return func(c *status.Cluster, n *status.Node) bool {
 		for i := 0; i < 5; i++ {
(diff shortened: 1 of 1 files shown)


pacoxu (commented):
/lgtm

since this is only used by node up-gradation.

neolit123:
let's merge this to let it run in CI.
/hold cancel

PR #2666, opened 2022-03-17
[WIP] Add a feature to specify alternate image repository in kinder

dims:
Signed-off-by: Davanum Srinivas <davanum@gmail.com>

### DIFF
--- a/kinder/cmd/kinder/create/cluster/createcluster.go
+++ b/kinder/cmd/kinder/create/cluster/createcluster.go
@@ -32,6 +32,7 @@ const (
 type flagpole struct {
 	Name                 string
 	ImageName            string
+	ImageRepository      string
 	Workers              int
 	ControlPlanes        int
 	Retain               bool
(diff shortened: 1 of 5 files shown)


neolit123 (commented):
@dims 

@ameukam joined the kubeadm office hours and we decided we are not going to do that.
the bulk of kubeadm used images comes from baking them from the CI GCS bucket tars. only coredns / etcd are pulled from the GCR.

in early 1.25 we can switch kubeadm proper to use the new registry which will also reflect in kinder tests.

in the meantime, @ameukam agreed to PR the new domain here:
https://github.com/kubernetes/kubeadm/blob/main/tests/e2e/manifests/verify_manifest_lists.go#L68
which will validate all required kubeadm manifests (without downloading layers) and in my book that is more that sufficient signal.

meeting notes / VOD:
https://docs.google.com/document/d/1ONcoy8bOw8SWPUwXxnKeRZST3lnUxYpUv4Y6466h9Ek/edit

dims:
ACK thanks for the update @neolit123 

/close

PR #2672, opened 2022-03-27
don't set conntrack parameters in kube-proxy

SataQiu:
It seems the kernel doesn't allow to set some conntrack fields
from non-init netns because they are global, so setting it in a
namespaces leaks it to other namespace:

netfilter: conntrack: Make global sysctls readonly in non-init netns
torvalds/linux@671c54e

By default kube-proxy tries to set nf_conntrack_max, that is readonly,
hence failing and the kproxy pods fail to start crashlooping.

Ref: kubernetes-sigs/kind#2241 kubernetes-sigs/kind#2240

### DIFF
--- a/kinder/pkg/kubeadm/config.go
+++ b/kinder/pkg/kubeadm/config.go
@@ -222,6 +222,10 @@ apiVersion: kubeproxy.config.k8s.io/v1alpha1
 kind: KubeProxyConfiguration
 metadata:
   name: config
+conntrack:
+  # Skip setting sysctl value "net.netfilter.nf_conntrack_max"
+  # It is a global variable that affects other namespaces
+  maxPerCore: 0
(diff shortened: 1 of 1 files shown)


neolit123 (commented):
/lgtm

given kind has merged a similar change, but i don't think we have been affected.
is there evidence for failures on the kinder side?

...i guess it might only affect users that test with kinder on a particular Linux host.

SataQiu:
> /lgtm
> 
> given kind has merged a similar change, but i don't think we have been affected. is there evidence for failures on the kinder e2e test side?
> 
> ...i guess it might only affect users that test with kinder on a particular Linux host.

Yes, I encountered this problem when testing Kinder locally. With this patch, Kinder can support more kernel versions.

Issue #2653, opened 2022-02-09
add dry run e2e tests


neolit123:
kubeadm is currently missing integration / e2e tests for `--dry-run`.
this means if we happen to break our dry run support for a particular command (e.g. `init`) we will not know about it until users report it to us.

xref https://github.com/kubernetes/kubeadm/issues/2649

kubeadm has integration tests here:
https://github.com/kubernetes/kubernetes/tree/master/cmd/kubeadm/test/cmd
these tests execute a precompiled kubeadm binary to perform some checks and look for exist status 0.

we can use the same method for the `init`, `join`, `reset` tests with `--dry-run`.
because the dry-run will be reentrant.

but we cannot use this method for `upgrade *` commands, because the `--dry-run` for upgrade expects an existing cluster.
- kubeconfig files in /etc/kubernetes/...
- a running kube-apiserver
- other running components...

to have everything in the same place we can add dry-run as part of kinder e2e test workflow:
https://github.com/kubernetes/kubeadm/tree/main/kinder/ci/workflows

the workflow can look like the following:
- allocate a kinder cluster with 1 node
- call `kubeadm init --dry-run` on it (add `--upload-certs` and other special flags, how to test external CA?)
- call` kubeadm join --dry-run` on it (add `--control-plane`,  `--certificate-key` and other flags?)
- call `kubeadm reset --dry-run` on it
- call `kubeadm init ...` to create an actual k8s node
- call `kubeadm upgrade apply --dry-run` to dry run the "primary node" upgrade of this node
- call `kubeadm upgrade node --dry-run` to dry run the "secondary node" upgrade of this node
- .. cleanup

tasks:
- [x] add initial e2e test job for dry-run
k/kubeadm: https://github.com/kubernetes/kubeadm/pull/2673
k/test-infra PR: https://github.com/kubernetes/test-infra/pull/25997
- [ ] in k/k, improve kubeadm's join to use a custom dry-run client as noted [here](https://github.com/kubernetes/kubeadm/issues/2653#issuecomment-1096496562)
TODO
- [ ] update e2e test jobs to have join dry-run as noted [here](https://github.com/kubernetes/kubeadm/issues/2653#issuecomment-1096496562)
TODO

SataQiu:
@neolit123  It seems that we cannot run `kubeadm join --dry-run` without an actual Kubernetes control-plane. :sweat_smile:

The `join` phase will try to fetch the `cluster-info` ConfigMap even though in dry-run mode.

[...]

PR #2688, opened 2022-04-26
kinder: ensure the container runtime is started before executing dryrun e2e test

SataQiu:
xref: https://github.com/kubernetes/kubeadm/issues/2653#issuecomment-1104535529


neolit123 (commented):
Thanks. This seems good to me. I will check again if it fixes the flakes in the next few days.

/area kinder
/kind failing-test
/lgtm

Issue #1682, opened 2019-07-22
Allow usage of KubeletConfiguration next to JoinConfiguration on kubeadm join


fabriziopandini:
as per @neolit123  [comment](https://github.com/kubernetes/enhancements/pull/1159#discussion_r305806740)

> we need to start respecting KubeletConfiguration next to JoinConfiguration on kubeadm join.
last time i checked we weren't.

related to this KEP:
https://github.com/kubernetes/enhancements/pull/1439

---------
1.25 tracking of kubeadm patch support for kubeletconfiguration:

- [x] KEP update
https://github.com/kubernetes/enhancements/pull/3312
- [x] k/k code PRs:
https://github.com/kubernetes/kubernetes/pull/110405
https://github.com/kubernetes/kubernetes/pull/110598
- [x] k/kubeadm test update
https://github.com/kubernetes/kubeadm/pull/2706
https://github.com/kubernetes/kubeadm/pull/2707
https://github.com/kubernetes/test-infra/pull/26557
- [x] docs update for https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/control-plane-flags/ https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/kubelet-integration/
https://github.com/kubernetes/website/pull/34259

fabriziopandini:
the tricky part here is how to preserve node-specific settings during upgrades...

rosti:
Allowing for customization on a per-node basis - yes, allowing for a replacement - probably no.
Component configs are shared on a per-cluster basis and are stored in config maps.
Allowing for them to be patched with node-local settings is a must have, but allowing for their total replacement is probably not desired.

fabriziopandini:
I'm +1 to merge this with the "Advanced configurations with kubeadm (Kustomize)" effort (customization on a per-node basis)

rosti:
My bet is to push in the Kustomize direction. I don't think, that there are many users who push for such kind of feature and, therefore, having the delay (necessary for Kustomize) is acceptable.

BenTheElder:
Without this you can customize kubelet per node but it requires using flags which is problematic given that many are deprecated.

fabriziopandini:
Rif https://github.com/kubernetes/enhancements/pull/1439

stealthybox:
> the tricky part here is how to preserve node-specific settings during upgrades...

^ the NodeGroup ConfigMap thing makes sense for persistent configurations.
[...]

PR #2706, opened 2022-06-08
kinder: make sure worker nodes also get a patches dir created

neolit123:
During join of worker nodes kinder always passes to kubeadm
JoinConfiguration that has a Patches.Directory.

If the directory does not exist kubeadm will error out.
Always create the directory for worker nodes.
This was already done for the "init" node and for "join" control
plane nodes.

fixes failing tests such as:
https://testgrid.k8s.io/sig-cluster-lifecycle-kubeadm#kubeadm-kinder-latest

xref https://github.com/kubernetes/kubeadm/issues/1682


neolit123:
cc @RA489 @pacoxu 

will self-LGTM this one. tested it locally and it works.

PR #2707, opened 2022-06-13
kinder: add test for patching kubeletconfiguration; cleanup test jobs

neolit123:
```
Add a kubeletconfiguration+strategic.yaml patch
that contains "shutdownGracePeriod: 1s".

Verify if the patch is applied. This is done only for
CP nodes, but exercises init/join/upgrade and the
patching / writing the kubelet config for worker nodes
is in the same code path.
```

```
We can only test patches at the k/k "master" level
to avoid having too many test jobs.

Ideally we should not backport "patches" fixes
to the old branches, unless they are already present
at k/k master.
```

xref https://github.com/kubernetes/kubeadm/issues/1682


RA489:
/lgtm
/hold for another eye @pacoxu 

feel free to cancel hold.

pacoxu (approved):
/hold cancel
/lgtm

neolit123:
looks like CI is failing for the new kubelet patch test:

> ERROR: /var/lib/kubelet/config.yaml is not patched
One or more verify checks failed! See output above...
Error: failed to exec command: failed to execute command on node kinder-patches-control-plane-1: exit status 1

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-kubeadm-kinder-patches-latest/1536760310143127552
https://testgrid.k8s.io/sig-cluster-lifecycle-kubeadm#kubeadm-kinder-patches-latest

i will look tomorrow.

pacoxu:
The log shows
```
[patches] Applied patch of type "application/strategic-merge-patch+json" to target "kubeletconfiguration"
[kubelet-start] Writing kubelet configuration to file "/var/lib/kubelet/config.yaml"
[kubelet-start] Starting the kubelet
```
It is really weird.

Issue #2671, opened 2022-03-24
switch the default kubeadm image registry to registry.k8s.io


neolit123:
the k8s project is moving away from k8s.gcr.io to registry.k8s.io.

- registry.k8s.io is currently redirecting to k8s.gcr.io
- in 1.25 we should switch kubeadm's code to use registry.k8s.io by default.
- k8s.gcr.io will continue working (for some time) by redirecting to registry.k8s.io

https://groups.google.com/g/kubernetes-sig-testing/c/U7b_im9vRrM/m/7qywJeUTBQAJ
xref https://github.com/kubernetes/k8s.io/issues/1834

1.25

- [x] migrate kubeadm to use the new default registry
https://github.com/kubernetes/kubernetes/pull/109938
- [x] mutate existing user default image registry to the new one
https://github.com/kubernetes/kubernetes/pull/110343
- [x] fix failing kinder / CI
https://github.com/kubernetes/kubeadm/pull/2705
- [x] update k/website pages
https://github.com/kubernetes/website/pull/34163

1.26 
- [x] cleanup [TODOs](https://github.com/kubernetes/kubernetes/pull/110343) from 1.25
https://github.com/kubernetes/kubernetes/pull/112006

- cherrypicks for 1.22, 1.23, 1.24:
https://github.com/kubernetes/kubernetes/pull/113388
https://github.com/kubernetes/kubernetes/pull/113393
https://github.com/kubernetes/kubernetes/pull/113395
(minus cleanups)

pacoxu:
https://github.com/kubernetes/kubernetes/pull/109532
https://github.com/kubernetes/kubernetes/pull/109533

neolit123:
PR for this in 1.25:
https://github.com/kubernetes/kubernetes/pull/109938

neolit123:
https://github.com/kubernetes/kubernetes/pull/109938 broke out e2e tests for 'latest'
https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-kubeadm

I think what happening is the following. A cluster is created with the older version of kubeadm and the clusterconfiguration.imageRepository is defaulted to "k8s gcr.io". The when upgrade happens the new kubeadm binary thinks that "k8s.gcr.io" is a custom repository. Kinder is not pre-pulling the right images.

For kubeadm we need to mutate the image repo field to registry.k8s.io.

For kinder we need to ensure it uses the ouput of "kubeadm config images" for prepull (might be the case already).

[...]

PR #2709, opened 2022-06-14
check image in registry.k8s.io instead of gcr

pacoxu:
xref https://github.com/kubernetes/kubeadm/issues/2671

as we have migrated to registry.k8s.io, should we check it instead?

### DIFF
--- a/tests/e2e/manifests/verify_manifest_lists.go
+++ b/tests/e2e/manifests/verify_manifest_lists.go
@@ -65,7 +65,7 @@ const (
 	// first release that is known to have full support.
 	firstKnownVersion = "v1.12.0-rc.1"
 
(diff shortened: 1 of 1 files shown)


neolit123 (commented):
@pacoxu why did you close this PR?
Wasn't it valid?

pacoxu:
> @pacoxu why did you close this PR? Wasn't it valid?

Before I opened the PR, the code in my local env was not updated. I found the registry URL was updated.
https://github.com/kubernetes/kubeadm/blob/e3b44af265702c458268afa6ef0ba9b3676eea85/tests/e2e/manifests/verify_manifest_lists.go#L68

Only the var name is not updated. I am not sure if it is needed.

PR #2744, opened 2022-08-16
update go format

pacoxu:
fix CI failure

### DIFF
--- a/kinder/pkg/cri/nodes/common/util.go
+++ b/kinder/pkg/cri/nodes/common/util.go
@@ -215,7 +215,7 @@ func ContainerArgsForExternalEtcd(name string, args []string) []string {
 	return args
 }
 
-// TryUntil implements an helper that calls `try()`` in a loop until the deadline `until`
+// TryUntil implements an helper that calls `try()` in a loop until the deadline `until`
 // has passed or `try()`returns true, returns whether try ever returned true
 func TryUntil(until time.Time, try func() bool) bool {
 	for until.After(time.Now()) {
--- a/kinder/ci/tools/update-workflows/templates/workflows/presubmit-upgrade-latest.yaml
+++ b/kinder/ci/tools/update-workflows/templates/workflows/presubmit-upgrade-latest.yaml
@@ -112,7 +112,6 @@ tasks:
     - test
     - e2e
     - --test-flags=--report-dir={{ .env.ARTIFACTS }} --report-prefix=e2e --ginkgo.dryRun=true
-    - --parallel
     - --name={{ .vars.clusterName }}
     - --loglevel=debug
   timeout: 35m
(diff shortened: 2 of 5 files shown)


kinder/pkg/cri/nodes/common/util.go
```diff
@@ -215,7 +215,7 @@ func ContainerArgsForExternalEtcd(name string, args []string) []string {
 	return args
 }
 
-// TryUntil implements an helper that calls `try()`` in a loop until the deadline `until`
+// TryUntil implements an helper that calls `try()“ in a loop until the deadline `until`
```
SataQiu, line 218:
```suggestion
// TryUntil implements an helper that calls `try()` in a loop until the deadline `until`
```

RA489: nice catch @SataQiu

RA489: nice catch @SataQiu

pacoxu: Fixed.

RA489: thanks

neolit123 (commented):
/approve

neolit123:
by the looks of things the new ginkgo cannot auto-ignore the -p flag when using dryrun (see the job spec for pull-kubeadm-kinder-upgrade-latest)  and we have and omit it.

somehow in kinder code / jobs we have to handle this.

> [1mOutput from proc 2:[0m
  Aug 16 05:02:05.701: INFO: The --provider flag is not set. Continuing as if --provider=skeleton had been used.
  I0816 05:02:05.701933   65630 e2e.go:116] Starting e2e run "965038c6-672c-4de0-a7f5-5971c567b904" on Ginkgo node 2
  [38;5;9mGinkgo detected configuration issues:[0m
  [1m[38;5;9mGinkgo only performs -dryRun in serial mode.[0m
    Please try running ginkgo -dryRun again, but without -p or -procs to ensure
    the suite is running in series.


or maybe there are more errors?

neolit123:
@pacoxu would you have time to implement the above -p modification. perhaps we can check the incomming list of flags from the workflow and if -dryrun is there we can drop -p?

if you cannot i can check before friday.

pacoxu:
> @pacoxu would you have time to implement the above -p modification. perhaps we can check the incomming list of flags from the workflow and if -dryrun is there we can drop -p?

I will check it next two days if I have time.

pacoxu:
Firstly, I tried to use GINKGO_PARALLEL instead of `--parallel` of testing.
However, the task does not support `env`. Then I removed the `--parallel`

neolit123 (commented):
/lgtm
/approve
/area kinder
/kind cleanup

i think this is ok, i did not see that we are adding --parallel explicitly

neolit123:
thanks @pacoxu

PR #2753, opened 2022-09-01
verify_manifest_list.go: support gzip and non-gzip layer media

neolit123:
The pause 3.8 image layers are of non-gzip TAR type, while
the test only supports gzip-ed layers. Handle both cases.

fixes failing tests:
> ERROR(cached result): pause:3.8; error: not a layer: application/vnd.docker.image.rootfs.diff.tar

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/periodic-kubernetes-e2e-manifest-lists/1565206029694668800
https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-all#periodic-manifest-lists


neolit123:
@RA489 @pacoxu PTAL for lgtm, thanks

PR #2840, opened 2023-03-15
Replace error: [export VIP=192.168.0.40'] 

natereid72:
Replace error: [export VIP=192.168.0.40']  with: [export VIP=192.168.0…40]


SataQiu:
thanks @natereid72 
/approve

Issue #2848, opened 2023-03-28
1.26 dryrun job failing


neolit123:
has been failing since 23th of march
https://testgrid.k8s.io/sig-cluster-lifecycle-kubeadm#kubeadm-kinder-dryrun-1-26

culprit seems to be upgrade where it tries 1.28, skipping 1.27

> --with-init-artifacts=v1.26.3-rc.0.57+e6323abe65c431 --with-upgrade-artifacts=v1.28.0-alpha.0.13+c3e7eca7fd3845

> Specified version to upgrade to "v1.28.0-alpha.0.13+c3e7eca7fd3845" is too high; kubeadm can upgrade only 1 minor version at a time

upgrade should be for 1.27. this hints the job fetches latest .28 artifacts.

pacoxu:
This is after the v1.28.0-alpha.0 released.

ci-kubernetes-e2e-kubeadm-kinder-upgrade-1-26-latest is using https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-e2e-kubeadm-kinder-upgrade-1-26-latest/1640598899607474176/build-log.txt
```
 kubeadm upgrade apply -f v1.28.0-alpha.0.13+c3e7eca7fd3845 --v=6
```

But ci-kubernetes-e2e-kubeadm-kinder-dryrun-1-26 is using without `-f`
```
/bin/bash -c docker exec kinder-dryrun-control-plane-1 /kinder/upgrade/v1.28.0-alpha.0.13+c3e7eca7fd3845/kubeadm upgrade apply v1.28.0-alpha.0.13+c3e7eca7fd3845 --ignore-preflight-errors=Swap,SystemVerification,FileContent--proc-sys-net-bridge-bridge-nf-call-iptables --allow-release-candidate-upgrades=true --allow-experimental-upgrades=true --dry-run=true --v=6 || exit 1

```
https://storage.googleapis.com/kubernetes-jenkins/logs/ci-kubernetes-e2e-kubeadm-kinder-dryrun-1-26/1640579773262467072/build-log.txt

SataQiu:
@neolit123 If we do them like the upgrade job, we will lose the init/join/reset dry-run tests for the latest version.
Thanks for your quick fix @pacoxu

PR #2850, opened 2023-03-28
add -f for kubeadm upgrade apply dryrun

pacoxu:
Fixes https://github.com/kubernetes/kubeadm/issues/2848

following 
https://github.com/kubernetes/kubeadm/blob/d6a85093203e5d6984339609e24ecb19bb1eb1e9/kinder/ci/tools/update-workflows/templates/workflows/upgrade-latest-no-addon-config-maps.yaml#L92-L93


neolit123 (commented):
-f can work but isn't it better to make it latest-1.27 produced by the config, as mentioned on the issue?

/lgtm
/approve
/hold

kinder/ci/workflows/skew-kubelet-1.23-on-1.24.yaml
```diff
@@ -0,0 +1,15 @@
+# AUTOGENERATED by https://git.k8s.io/kubeadm/kinder/ci/tools/update-workflows
```
SataQiu, line 1: This does not seem to be about dryrun？

SataQiu (commented):
/lgtm
/approve

kinder/ci/workflows/skew-kubelet-1.23-on-1.24.yaml
pacoxu: Not familiar with the `./hack/update-workflows.sh` script, and the v1.23 related skew are generated. Is there a flag or place to configure it?

pacoxu:
> -f can work but isn't it better to make it latest-1.27 produced by the config, as mentioned on the issue?

As there is a new tag v1.28.0-alpha.0 in k/k, I am not sure if we can switch back to make latest-v1.27 upgrade automatically.

SataQiu:
After https://github.com/kubernetes/test-infra/pull/29137 merged, we can update our ci workflows for v1.27.

neolit123:
> this fix should be valid.

alright, cancel hold when needed.

SataQiu:
@pacoxu How about squashing the two commits into one, and then merging this PR?

pacoxu:
> @pacoxu How about squashing the two commits into one, and then merging this PR?

Squashed.

Issue #2873, opened 2023-05-12
fix broken manifest list e2e due to arm arch


neolit123:
looks like the e2e test job has been broken for a while
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/periodic-kubernetes-e2e-manifest-lists/1656908203142680576

reading the logs shows missing arm arch for images

> I0512 06:25:22.591] * ERROR: kube-apiserver:v1.27.1; error: did not find a match for these architectures: arm
I0512 06:25:22.591] 

is arm no longer supported? if yes, we need to update the test code.

if no, we need to log k/k issue ad ping #sig-release

NOTE also it would be nice to make this e2e job send email alerts. it currently does not!
https://github.com/kubernetes/kubeadm/pull/2872#discussion_r1192555102

dims:
> is arm no longer supported?

yes, let's update the test code!

BenTheElder:
Yes, https://github.com/kubernetes/kubernetes/pull/115742

We have a build time warning throughout the 1.27 branch, I dropped it for 1.28 development branch (master) after release.

If you build 1.27.x:
```
+++ [0512 19:03:40] Running build command...
+++ [0512 19:03:40] WARNING: linux/arm will no longer be built/shipped by default, please build it explicitly if needed.
+++ [0512 19:03:40]          support for linux/arm will be removed in a subsequent release.
+++ [0512 19:03:40] Building go targets for linux/amd64
```

rjsadow:
https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-all#periodic-manifest-lists test is passing again.

chendave:
Can we close this issue now?

pacoxu:
/close
as the ci is green

PR #2874, opened 2023-05-12
fix: Add support for removing unsupported architectures 

rjsadow:
This PR introduces a map called `architecturesToRemove` to `verify_manifest_list.go`, which maps the first unsupported version of Kubernetes to a list of architectures that should be removed. In the function `verifyManifestList`, the unsupported architectures are removed from the architecture list based on the Kubernetes version provided. This ensures that only supported architectures are considered during verification.

/kind bug
ref: https://github.com/kubernetes/kubeadm/issues/2873

### DIFF
--- a/tests/e2e/manifests/verify_manifest_lists.go
+++ b/tests/e2e/manifests/verify_manifest_lists.go
@@ -106,6 +106,10 @@ var (
 	// status of images is cached here, so that the same image is not
 	// tested by multiple tests.
 	verifiedImageCache = make(map[string]error)
+	// define a map where the keys are the first unseported version and the values are slices of architectures to be removed
+	architecturesToRemove = map[string][]string{
+		"1.27.0-beta.0": []string{"arm"},
+	}
 )
 
 // bellow are some types as per the docker specs.
@@ -444,7 +448,7 @@ func verifyArchImage(arch, imageName, archImage string) error {
 }
 
 // verify a manifest list and match the required architectures.
-func verifyManifestList(manifest, imageName, tag string) error {
+func verifyManifestList(manifest, imageName, tag string, ver *version.Version) error {
 	ml := manifestList{}
 	if err := json.Unmarshal([]byte(manifest), &ml); err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


tests/e2e/manifests/verify_manifest_lists.go
```diff
@@ -459,6 +463,26 @@ func verifyManifestList(manifest, imageName, tag string) error {
 	aList := make([]string, len(archList))
 	// copy into a temp slice.
 	copy(aList, archList)
+	// remove arch's from the list if the k8s version no longer supports them
+	for versionToRemove, archs := range architecturesToRemove {
+		verWithoutPreRelease, err := version.ParseSemantic(fmt.Sprintf("%d.%d.%d", ver.Major(), ver.Minor(), ver.Patch()))
+		if err != nil {
+			return fmt.Errorf("could not parse version %q: %v", ver.String(), err)
+		}
+		if verWithoutPreRelease.AtLeast(version.MustParseSemantic(versionToRemove)) {
```
neolit123, line 472: is this correct? reading the logs it seems the arm arch was removed in 1.27 pre-release, but i think with this we will only not check for arm in >= 1.27.0.

rjsadow: `Version.AtLeast()` doesnt support evaluating pre releases which is why I created the `verWithoutPreRelease` variable. This way prerelease versions will be removed from semantic versioning so "1.27.0-rc.1" is evaluated as "1.27.0".

rjsadow: One last bit of confirmation: https://go.dev/play/p/gLsU3Oesu6C

rjsadow: That being said, because of the `AtLeast()` limitation, this code as is would not handle a situation where an arch was removed during a pre-release i.e supported in "1.X.0-rc.1" and removed in "1.X.0-rc.2". I think the chances of such a large change/decision happening between pre-release's is unlikely, but something to be aware of.

neolit123: >  I think the chances of such a large change/decision happening between pre-release's is unlikely, but something to be aware of.

it might happen, as a forgotten task before a release.
can we just use the Compare function with an exact version even if it's pre-release?

rjsadow: Awesome recommendation! It looks like that'll work. https://go.dev/play/p/fJF35HOtd53 I'll make the updates.

rjsadow: Ok, updated to work with prerelease versions. Thanks!

neolit123 (commented):
/lgtm
/approve
thanks!

Issue #2346, opened 2020-11-09
mutable upgrades: upgrade the kube-proxy DS after all control-plane components


neolit123:
during mutable upgrades, upgrade kube-proxy only after the kube-apiserver on all nodes have been upgraded.

currently kubeadm does the following:
- upgrades the control-plane on one node (via "apply")
- upgrades kube-proxy and coredns
- upgrades the rest of the control-plane nodes and worker nodes
- upgrades kubelets everywhere

the problem with that is that it can create issues for a period of time when kube-proxy's version is ahead of a kube-apiserver version in an HA setup.

https://kubernetes.io/docs/setup/release/version-skew-policy/#kube-proxy

there is also another problem where kube-proxy would be updated before the kubelets on nodes and the policy suggests they should match.

this would require:
- "apply" phases so that upgrades of the addons are performed last after kube-* control-plane components are upgraded:
https://github.com/kubernetes/kubeadm/issues/1318
- changes in the upgrade documentation:
https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/

TODO List:
- [x] introduce a new feature gate UpgradeAddonsBeforeControlPlane to fix a kube-proxy skew policy misalignment 
- https://github.com/kubernetes/kubernetes/pull/116570 
- https://github.com/kubernetes/kubernetes/pull/117660
- [x] add e2e tests for latest k8s release  
- https://github.com/kubernetes/kubeadm/pull/2884
- [x] update docs for alpha FG:
        https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade/
        https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-init/#feature-gates
- https://github.com/kubernetes/website/pull/41606
- [x] remove the deprecated featuregate - perhaps next release 1.31?
- https://github.com/kubernetes/kubernetes/pull/124715
- https://github.com/kubernetes/test-infra/pull/32599
- https://github.com/kubernetes/website/pull/46274

neolit123:
cc @fabriziopandini

yvespp:
Don't know how hard it would be but could the proxy maybe be converted to a static pod manifest to solve this?
The update of it could then be moved from `kubeadm upgrade apply <vesions>` on the master to `kubeadm upgrade node`.
1. Remove manifest of proxy to stop it
1. Stop kubelet
1. Update config and manifest
1. Start kubelet

[...]

PR #2884, opened 2023-05-30
Add e2e test for UpgradeAddonsBeforeControlPlane feature gate

SataQiu:
Ref #2346

test-infra PR: https://github.com/kubernetes/test-infra/pull/29628


neolit123 (commented):
/lgtm
i think this is fine
thanks

neolit123:
@SataQiu the upgrade part of the job is failing.
i did not investigate:
https://testgrid.k8s.io/sig-cluster-lifecycle-kubeadm#kubeadm-kinder-upgrade-addons-before-controlplane-1-27-latest

pacoxu:
Let me take a look.
It seems the `kubeadm upgrade ` is using the old version kubeadm. I need to do a deeper investigation.

kinder/pkg/cluster/manager/actions/kubeadm-upgrade.go
```diff
@@ -131,6 +131,9 @@ func kubeadmUpgradeApply(c *status.Cluster, cp1 *status.Node, upgradeVersion *K8
 			applyArgs = append(applyArgs, "--patches", constants.PatchesDir)
 		}
 	}
+	if len(featureGates) > 0 {
+		applyArgs = append(applyArgs, fmt.Sprintf("--feature-gates=%s", featureGates))
```
pacoxu, line 135: Success
```
 ./kubeadm upgrade apply -f v1.28.0-alpha.1.25+627f43d8bb6ce3 --v=6 --feature-gates="UpgradeAddonsBeforeControlPlane=true"
```

Failed with the same error
```
./kubeadm upgrade apply -f v1.28.0-alpha.1.25+627f43d8bb6ce3 --v=6 --feature-gates=\"UpgradeAddonsBeforeControlPlane=true\"
```

SataQiu: Why does escaping happen here?

SataQiu: I know the resaon, we need to do this:
```
        // We remove the leading and trailing double or single quotes because the
	// feature-gate could be set as
	// --kubeadm-feature-gate="RootlessControlPlane=true" or
	// --kubeadm-feature-gate='RootlessControlPlane=true', so the value
	// of featureGate string would be "\"RootlessControlPlane=true"\" or "RootlessControlPlane=true'" respectively.
	// Once we trim the value double or single quotes the value will be "RootlessControlPlane=true".
	trimmedFeatureGate := strings.Trim(featureGate, "\"'")
```

SataQiu:
Let's wait and see if CI turns green.

SataQiu:
> the upgrade part of the job is failing. i did not investigate: https://testgrid.k8s.io/sig-cluster-lifecycle-kubeadm#kubeadm-kinder-upgrade-addons-before-controlplane-1-27-latest

It's green now. thanks @neolit123 @pacoxu

PR #2886, opened 2023-06-01
Fix a bug in UpgradeAddonsBeforeControlPlane feature gate e2e test

SataQiu:
Ref #2884

### DIFF
--- a/kinder/pkg/cluster/manager/actions/kubeadm-upgrade.go
+++ b/kinder/pkg/cluster/manager/actions/kubeadm-upgrade.go
@@ -35,7 +35,7 @@ import (
 //
 // The implementation assumes that the kubeadm/kubelet/kubectl binaries and all the necessary images
 // for the new kubernetes version are available in the /kinder/upgrade/{version} folder.
-func KubeadmUpgrade(c *status.Cluster, upgradeVersion *K8sVersion.Version, patchesDir string, featureGates string, wait time.Duration, vLevel int) (err error) {
+func KubeadmUpgrade(c *status.Cluster, upgradeVersion *K8sVersion.Version, patchesDir string, featureGate string, wait time.Duration, vLevel int) (err error) {
 	if upgradeVersion == nil {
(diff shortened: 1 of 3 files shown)


kinder/pkg/cluster/manager/actions/kubeadm-upgrade.go
```diff
@@ -120,7 +120,7 @@ func upgradeKubeadmBinary(n *status.Node, upgradeVersion *K8sVersion.Version) er
 	return nil
 }
 
-func kubeadmUpgradeApply(c *status.Cluster, cp1 *status.Node, upgradeVersion *K8sVersion.Version, patchesDir string, featureGates string, wait time.Duration, vLevel int) error {
+func kubeadmUpgradeApply(c *status.Cluster, cp1 *status.Node, upgradeVersion *K8sVersion.Version, patchesDir string, featureGate string, wait time.Duration, vLevel int) error {
```
pacoxu, line 123: Overall LGTM.

why change featureGates to featureGate? It is confusing.

SataQiu: We can only support a single kubeadm feature-gate now.

SataQiu: // FeatureGate option sets **a single kubeadm feature-gate** for the kubeadm commands

SataQiu: For example: https://github.com/kubernetes/kubeadm/blob/2f7c488859fb8c75b4037903dc7895bf05ccec1c/kinder/pkg/cluster/manager/actions/kubeadm-config.go#L97-L100

pacoxu: Sorry, It seems to be limited at the beginning.

https://github.com/kubernetes/kubeadm/blob/2f7c488859fb8c75b4037903dc7895bf05ccec1c/kinder/cmd/kinder/do/do.go#L126-L130

PR #2939, opened 2023-09-29
Make checking api server script to detect http code

wb14123:
Improve the script to check api server in keepalived config document, so that it can check http code as well. For example, sometimes if api server is stuck on exit, curl will get http code 503 instead of exit with an error code.

### DIFF
--- a/docs/ha-considerations.md
+++ b/docs/ha-considerations.md
@@ -83,7 +83,7 @@ There are some placeholders in `bash` variable style to fill in:
 
 The above `keepalived` configuration uses a health check script `/etc/keepalived/check_apiserver.sh` responsible for making sure that on the node holding the virtual IP the API Server is available. This script could look like this:
 
-```
+```bash
 #!/bin/sh
 
 errorExit() {
@@ -91,15 +91,24 @@ errorExit() {
     exit 1
 }
 
-curl --silent --max-time 2 --insecure https://localhost:${APISERVER_DEST_PORT}/ -o /dev/null || errorExit "Error GET https://localhost:${APISERVER_DEST_PORT}/"
-if ip addr | grep -q ${APISERVER_VIP}; then
-    curl --silent --max-time 2 --insecure https://${APISERVER_VIP}:${APISERVER_DEST_PORT}/ -o /dev/null || errorExit "Error GET https://${APISERVER_VIP}:${APISERVER_DEST_PORT}/"
+# replace this with the virtual IP address negotiated between the `keepalived` cluster hosts
+APISERVER_VIP=${APISERVER_VIP_VALUE}
+# replace this with the port through which Kubernetes will talk to the API Server
+APISERVER_DEST_PORT=${APISERVER_DEST_PORT_VALUE}
+
+url="https://localhost:$APISERVER_DEST_PORT"
+if ip addr | grep -q $APISERVER_VIP; then
+        url="https://$APISERVER_VIP:$APISERVER_DEST_PORT"
 fi
+http_code=$(curl --output /dev/null --silent --write-out "%{http_code}" --insecure --max-time 2 ${url})
+err_code=$?
+[ $err_code -eq 0 ] || errorExit "Error get $url, error code $err_code"
+[ "$http_code" = "200" ] || [ "$http_code" = "403" ] || errorExit "Error get $url, status code $http_code"
 ```
 
 There are some placeholders in `bash` variable style to fill in:
-- `${APISERVER_VIP}` is the virtual IP address negotiated between the `keepalived` cluster hosts.
-- `${APISERVER_DEST_PORT}` the port through which Kubernetes will talk to the API Server.
+- `${APISERVER_VIP_VALUE}` is the virtual IP address negotiated between the `keepalived` cluster hosts.
+- `${APISERVER_DEST_PORT_VALUE}` the port through which Kubernetes will talk to the API Server.
 
 ### haproxy configuration
 


docs/ha-considerations.md
```diff
@@ -91,15 +91,24 @@ errorExit() {
     exit 1
 }
 
-curl --silent --max-time 2 --insecure https://localhost:${APISERVER_DEST_PORT}/ -o /dev/null || errorExit "Error GET https://localhost:${APISERVER_DEST_PORT}/"
-if ip addr | grep -q ${APISERVER_VIP}; then
-    curl --silent --max-time 2 --insecure https://${APISERVER_VIP}:${APISERVER_DEST_PORT}/ -o /dev/null || errorExit "Error GET https://${APISERVER_VIP}:${APISERVER_DEST_PORT}/"
+# replace this with the virtual IP address negotiated between the `keepalived` cluster hosts
+APISERVER_VIP="10.88.111.88"
```
neolit123, line 95: we shouldn't define these vars here

docs/ha-considerations.md
```diff
@@ -83,23 +83,32 @@ There are some placeholders in `bash` variable style to fill in:
...
+err_code=$?
+[[ $err_code == 0 ]] || errorExit "Error get $url, error code $err_code"
+[[ $http_code == "200" ]] || [[ $http_code == "403" ]] || errorExit "Error get $url, status code $http_code"
 ```
 
 There are some placeholders in `bash` variable style to fill in:
-- `${APISERVER_VIP}` is the virtual IP address negotiated between the `keepalived` cluster hosts.
-- `${APISERVER_DEST_PORT}` the port through which Kubernetes will talk to the API Server.
+- `APISERVER_VIP` is the virtual IP address negotiated between the `keepalived` cluster hosts.
+- `APISERVER_DEST_PORT` the port through which Kubernetes will talk to the API Server.
```
neolit123, line 111: it's better to reference with '${...}'.
you will see that shellcheck may complain about this.

docs/ha-considerations.md
```diff
@@ -83,23 +83,32 @@ There are some placeholders in `bash` variable style to fill in:
...
+# replace this with the virtual IP address negotiated between the `keepalived` cluster hosts
+APISERVER_VIP="10.88.111.88"
+# replace this with the port through which Kubernetes will talk to the API Server
+APISERVER_DEST_PORT="6443"
+
+url="https://localhost:$APISERVER_DEST_PORT"
+if ip addr | grep -q $APISERVER_VIP; then
+        url="https://$APISERVER_VIP:$APISERVER_DEST_PORT"
 fi
+http_code=`curl --output /dev/null --silent --write-out "%{http_code}" --insecure --max-time 2 ${url}`
```
neolit123, line 103: this is not a portable command execution and result assignment

neolit123 (commented):
/cc @mbert 
PTAL if you have a moment

docs/ha-considerations.md
wb14123: It is intended to assigned the result to `http_code` so that we can check it later. Which part of it is not portable?

neolit123: see https://www.shellcheck.net/

docs/ha-considerations.md
wb14123: It's defined so that it can be reused in the script. If you mean the value of `"10.88.111.88"`, I changed the code to put `${APISERVER_VIP_VALUE}` there so that we don't have a random value there.

docs/ha-considerations.md
wb14123: Fixed

docs/ha-considerations.md
wb14123: Fixed the warnings on there

docs/ha-considerations.md
```diff
@@ -83,23 +83,32 @@ There are some placeholders in `bash` variable style to fill in:
...
 }
 
-curl --silent --max-time 2 --insecure https://localhost:${APISERVER_DEST_PORT}/ -o /dev/null || errorExit "Error GET https://localhost:${APISERVER_DEST_PORT}/"
-if ip addr | grep -q ${APISERVER_VIP}; then
-    curl --silent --max-time 2 --insecure https://${APISERVER_VIP}:${APISERVER_DEST_PORT}/ -o /dev/null || errorExit "Error GET https://${APISERVER_VIP}:${APISERVER_DEST_PORT}/"
+# replace this with the virtual IP address negotiated between the `keepalived` cluster hosts
+APISERVER_VIP=${APISERVER_VIP_VALUE}
+# replace this with the port through which Kubernetes will talk to the API Server
+APISERVER_DEST_PORT=${APISERVER_DEST_PORT_VALUE}
+
```
neolit123, line 98: there is no point having both VALUE and the non- VALUE variables.

docs/ha-considerations.md
```diff
@@ -83,23 +83,32 @@ There are some placeholders in `bash` variable style to fill in:
...
+err_code=$?
+[ $err_code -eq 0 ] || errorExit "Error get $url, error code $err_code"
+[ "$http_code" = "200" ] || [ "$http_code" = "403" ] || errorExit "Error get $url, status code $http_code"
 ```
 
 There are some placeholders in `bash` variable style to fill in:
-- `${APISERVER_VIP}` is the virtual IP address negotiated between the `keepalived` cluster hosts.
-- `${APISERVER_DEST_PORT}` the port through which Kubernetes will talk to the API Server.
+- `${APISERVER_VIP_VALUE}` is the virtual IP address negotiated between the `keepalived` cluster hosts.
+- `${APISERVER_DEST_PORT_VALUE}` the port through which Kubernetes will talk to the API Server.
```
neolit123, line 111: the vars were listed here without _VALUE.
this change is not needed

docs/ha-considerations.md
```diff
@@ -83,23 +83,32 @@ There are some placeholders in `bash` variable style to fill in:
...
 
-curl --silent --max-time 2 --insecure https://localhost:${APISERVER_DEST_PORT}/ -o /dev/null || errorExit "Error GET https://localhost:${APISERVER_DEST_PORT}/"
-if ip addr | grep -q ${APISERVER_VIP}; then
-    curl --silent --max-time 2 --insecure https://${APISERVER_VIP}:${APISERVER_DEST_PORT}/ -o /dev/null || errorExit "Error GET https://${APISERVER_VIP}:${APISERVER_DEST_PORT}/"
+# replace this with the virtual IP address negotiated between the `keepalived` cluster hosts
+APISERVER_VIP=${APISERVER_VIP_VALUE}
+# replace this with the port through which Kubernetes will talk to the API Server
+APISERVER_DEST_PORT=${APISERVER_DEST_PORT_VALUE}
+
+url="https://localhost:$APISERVER_DEST_PORT"
```
neolit123, line 99: uppercase URL

docs/ha-considerations.md
```diff
@@ -83,23 +83,32 @@ There are some placeholders in `bash` variable style to fill in:
...
+if ip addr | grep -q $APISERVER_VIP; then
+        url="https://$APISERVER_VIP:$APISERVER_DEST_PORT"
 fi
+http_code=$(curl --output /dev/null --silent --write-out "%{http_code}" --insecure --max-time 2 ${url})
```
neolit123, line 103: uppercase HTTP_CODE
what does --write-out %... do here?

docs/ha-considerations.md
```diff
@@ -83,23 +83,32 @@ There are some placeholders in `bash` variable style to fill in:
...
+err_code=$?
```
neolit123, line 104: uppercase ERR_CODE

docs/ha-considerations.md
wb14123: You mean delete `APISERVER_VIP`? It's needed in the script below. If you want to put `APISERVER_VIP` in the script below and let user replace that string, then user need to replace it multiple times. So why not declare it here so that user can just replace the value of if once?

docs/ha-considerations.md
wb14123: Does kubeadm has a different code style for bash script? Normally local non constant vars should be lower case.

docs/ha-considerations.md
wb14123: `--write-out "%{http_code}"` print out the http status code

docs/ha-considerations.md
neolit123: remove *_VALUE vars.

docs/ha-considerations.md
neolit123: what is the rest of the doc doing?

wb14123: I don't see any other scripts in this doc. If the naming practice is different than the normal ones, just tell me so, or better, put code style guideline somewhere instead of expect people to go through all the docs and find random scripts to see what is the "consistent" style.

neolit123: there is no convention documented in this repo or kubernetes/kubernetes, but uppercase even for local vars seems more common.

wb14123: Almost all the shell code styles I saw use lowercase in this case, including [Google's shell style guide](https://google.github.io/styleguide/shellguide.html#variable-names)

docs/ha-considerations.md
mbert: The err_code variable is unnecessary. It's really cosmetical, but I'd probably prefer to see the code structured like this:
```
HTTP_CODE=$(curl --output /dev/null --silent --write-out "%{http_code}" --insecure --max-time 2 ${URL}) || errorExit "Error get '$URL'."

# some words explaining why code 403 is considered OK
test "$HTTP_CODE" = "200" -o "$HTTP_CODE" = "403" || errorExit "Error get '$URL', status code '$HTTP_CODE'."
```

There's no need to track a nonzero exit code, hence no need to keep it in a variable and add an additional line for checking it.

There's no need for `[ ... ] || [ ... ]` constructions, an OR expression is something a single `test` command is perfectly capable of (I personally prefer to write `test` rather than `[ ... ]`, but that's a matter of taste.

Whenever quoting status codes and the like in shell scripts using single quotes is recommendable for better readability of the output and making spelling errors (resulting in empty variable values) easier to find.

mbert (changes requested):

neolit123 (commented):
@mbert thanks!
@wb14123 do you object to the requested changes?

docs/ha-considerations.md
chendave: sorry, I didn't get it, why we need this placeholders? shouldn't the env var "APISERVER_VIP" and "APISERVER_DEST_PORT" are enough, they can be replaced by the actual IP addr or port, right?

docs/ha-considerations.md
```diff
@@ -83,23 +83,32 @@ There are some placeholders in `bash` variable style to fill in:
...
+[ $err_code -eq 0 ] || errorExit "Error get $url, error code $err_code"
+[ "$http_code" = "200" ] || [ "$http_code" = "403" ] || errorExit "Error get $url, status code $http_code"
```
chendave, line 106: echo to the comment above, comment here to explain why "403" is not considered as an error code.

mbert: @chendave regarding the placeholders: in my original example script there were no variables for `APISERVER_VIP` and `APISERVER_DEST_PORT`, hence these names were placeholders to be filled in when creating the script for a particular network. 

@wb14123 has changed the script a bit, so that now actual variables are used. Hence the above names were used for variables in the script and new names, `APISERVER_VIP_VALUE` and `APISERVER_DEST_PORT_VALUE` were introduced as placeholders instead. Again, these are not meant as "real" variables but values to fill in when creating the script for a network. 

I hope this is understandable now?

neolit123:
closing as inactive. please reopen if you wish to continue with the updates here.

Issue #2924, opened 2023-09-04
adjust the kubeadm / kubelet skew policy


NorthFuture:
edit by neolit123

action items:
- [x] adjust all kubelet skew constants / checks for init/join/upgrade.
https://github.com/kubernetes/kubernetes/pull/120825
- [x] update the "create a cluster with kubeadm" page where our skew is documented. it must be for the release-1.29 k/website branch.
https://github.com/kubernetes/website/pull/43769
- [ ] add the -f flag for `upgrade node`. this is a nice to have, we missed 1.29, so it can be added in 1.30.
TODO
- [ ] add one new e2e test, that upgrades with the new kubelet skew without -f.
(this is actually difficult because -f is required in our CI;  it's used for kubeadm to allow upgrading to a pre-release / CI artifact)
TODO
- [ ] in 1.32 remove `--ignore-preflight-errors=KubeletVersion` from the kinder kubelet skew jobs [see this note for details](https://github.com/kubernetes/kubeadm/pull/2944#pullrequestreview-1726883980)
TODO

how was 1.32 established:
```
kubeadm 1.29 is the first release that supports the new skew
kubeadm 1.29 supports deploying kubelet 1.29, 1.28, 1.27, 1.26
k8s (kubeadm) support window is 3x releases at a time
target kubeadm version to drop the preflight = when 1.28 goes out of support / 1.32 is released.
note: the kubeadm/CP skew is ignored in this case even if the kubelet/CP skew is the actual target of this change
```

----
FEATURE REQUEST:

with kubernetes 1.28 the skew policy for control plane components has been updated, and now you can have control plane components that's three versions ahead than kubelets. 

However kubeadm has still a n-1 skew policy. Such policy prevents to skip some kubelet upgrates of worker nodes, that could save a lot of time during upgrades in large clusters.

edit(neolit123): KEP LINK:
`Support Oldest Node And Newest Control Plane`
https://github.com/kubernetes/enhancements/tree/master/keps/sig-architecture/3935-oldest-node-newest-control-plane

## Versions

**kubeadm version** (use `kubeadm version`): 1.28

**Environment**:
- **Kubernetes version** (use `kubectl version`): 1.28
- **Cloud provider or hardware configuration**: N/A
- **OS** (e.g. from /etc/os-release): N/A
- **Kernel** (e.g. `uname -a`): N/A
- **Container runtime (CRI)** (e.g. containerd, cri-o): N/A
[...]

PR #2944, opened 2023-10-18
[wip]remove the KubeletVersion from ignorePreflightErrors for v1.29

pacoxu:
/cc @neolit123 @lalitc375

xref https://github.com/kubernetes/kubeadm/issues/2924

### DIFF
--- a/kinder/ci/tools/update-workflows/config.yaml
+++ b/kinder/ci/tools/update-workflows/config.yaml
@@ -88,21 +88,27 @@ jobGroups:
   - kubernetesVersion: 0
     kubeletVersion: -1
     skipVersions: [ 0 ]
+    ignorePreflightErrors: KubeletVersion
   - kubernetesVersion: 0
     kubeletVersion: -2
     skipVersions: [ -1, 0 ]
+    ignorePreflightErrors: KubeletVersion
   - kubernetesVersion: 0
     kubeletVersion: -3
     skipVersions: [ -2, -1, 0 ]
+    ignorePreflightErrors: KubeletVersion
   - kubernetesVersion: -1
     kubeletVersion: -2
     skipVersions: [ -1 ]
+    ignorePreflightErrors: KubeletVersion
   - kubernetesVersion: -1
     kubeletVersion: -3
     skipVersions: [ -2, -1 ]
+    ignorePreflightErrors: KubeletVersion
   - kubernetesVersion: -2
     kubeletVersion: -3
     skipVersions: [ -2 ]
+    ignorePreflightErrors: KubeletVersion
 
 - name: skew-x-on-y
   testInfraJobSpec:
@@ -118,10 +124,13 @@ jobGroups:
     kubeadmVersion: latest
   - kubernetesVersion: -1
     kubeadmVersion: 0
+    ignorePreflightErrors: KubeletVersion
   - kubernetesVersion: -2
     kubeadmVersion: -1
+    ignorePreflightErrors: KubeletVersion
   - kubernetesVersion: -3
     kubeadmVersion: -2
+    ignorePreflightErrors: KubeletVersion
 
 - name: upgrade
   testInfraJobSpec:
--- a/kinder/ci/tools/update-workflows/templates/workflows/skew-kubelet-x-on-y.yaml
+++ b/kinder/ci/tools/update-workflows/templates/workflows/skew-kubelet-x-on-y.yaml
@@ -7,7 +7,7 @@ vars:
   kubeadmVersion: "\{\{ resolve `ci/{{ ciLabelFor .KubeadmVersion }}` \}\}"
   kubeletVersion: "\{\{ resolve `ci/{{ ciLabelFor .KubeletVersion }}` \}\}"
   kubernetesVersion: "\{\{ resolve `ci/{{ ciLabelFor .KubernetesVersion }}` \}\}"
-  ignorePreflightErrors: "KubeletVersion"
+  ignorePreflightErrors: "{{.IgnorePreflightErrors}}"
   ginkgoSkip: "\\[MinimumKubeletVersion:({{ .SkipVersions }})\\]"
   controlPlaneNodes: 3
 tasks:
--- a/kinder/ci/tools/update-workflows/templates/workflows/skew-x-on-y.yaml
+++ b/kinder/ci/tools/update-workflows/templates/workflows/skew-x-on-y.yaml
@@ -7,7 +7,7 @@ vars:
   kubeadmVersion: "\{\{ resolve `ci/{{ ciLabelFor .KubeadmVersion }}` \}\}"
   kubeletVersion: "\{\{ resolve `ci/{{ ciLabelFor .KubeletVersion }}` \}\}"
   kubernetesVersion: "\{\{ resolve `ci/{{ ciLabelFor .KubernetesVersion }}` \}\}"
-  ignorePreflightErrors: "KubeletVersion"
+  ignorePreflightErrors: "{{.IgnorePreflightErrors}}"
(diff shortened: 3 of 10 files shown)


kinder/ci/tools/update-workflows/templates/workflows/skew-kubelet-x-on-y.yaml
```diff
@@ -7,7 +7,7 @@ vars:
   kubeadmVersion: "\{\{ resolve `ci/{{ ciLabelFor .KubeadmVersion }}` \}\}"
   kubeletVersion: "\{\{ resolve `ci/{{ ciLabelFor .KubeletVersion }}` \}\}"
   kubernetesVersion: "\{\{ resolve `ci/{{ ciLabelFor .KubernetesVersion }}` \}\}"
-  ignorePreflightErrors: "KubeletVersion"
+  ignorePreflightErrors: "{{.IgnorePreflightErrors}}"
```
neolit123, line 10: i just realized this forces us to update the tools/update-workflows/config.yaml file on every new release, because the versions will shift. this is noisy / problematic.

how about we keep the `ignorePreflightErrors: "KubeletVersion"` hardcoded for all kubelet skews until we no longer need it ~3x releases from now?

i think unit tests in k/k should be enough to verify we support the N-3 skew, right?

kinder/ci/tools/update-workflows/config.yaml
```diff
@@ -118,10 +124,13 @@ jobGroups:
     kubeadmVersion: latest
   - kubernetesVersion: -1
     kubeadmVersion: 0
+    ignorePreflightErrors: KubeletVersion
   - kubernetesVersion: -2
     kubeadmVersion: -1
+    ignorePreflightErrors: KubeletVersion
   - kubernetesVersion: -3
     kubeadmVersion: -2
+    ignorePreflightErrors: KubeletVersion
```
neolit123, line 133: `ignorePreflightErrors: KubeletVersion` is not needed, because the kubeletversion == kubernetesversion?

kinder/ci/tools/update-workflows/templates/workflows/skew-x-on-y.yaml
```diff
@@ -7,7 +7,7 @@ vars:
   kubeadmVersion: "\{\{ resolve `ci/{{ ciLabelFor .KubeadmVersion }}` \}\}"
   kubeletVersion: "\{\{ resolve `ci/{{ ciLabelFor .KubeletVersion }}` \}\}"
   kubernetesVersion: "\{\{ resolve `ci/{{ ciLabelFor .KubernetesVersion }}` \}\}"
-  ignorePreflightErrors: "KubeletVersion"
+  ignorePreflightErrors: "{{.IgnorePreflightErrors}}"
```
neolit123, line 10: can we just remove this old line:
> ignorePreflightErrors: "KubeletVersion"

i don't remember why we added it.
the kubelet version should match the k8s version and the k8s version is N-1 of the kubeadm version.

neolit123 (commented):
/hold for review

neolit123 (commented):
what is the status of this @pacoxu ?
if you don't have time i can send the test / docs prs?

kinder/ci/tools/update-workflows/templates/workflows/skew-kubelet-x-on-y.yaml
pacoxu: If so, we will ignore all KubeletVersion in the skew test here.

This will be easier to maintain. (The current change is more accurate to me. 🤣)

> i think unit tests in k/k should be enough to verify we support the N-3 skew, right?

Yes. This can be covered IIRC

kinder/ci/tools/update-workflows/templates/workflows/skew-x-on-y.yaml
pacoxu: https://github.com/kubernetes/kubeadm/blob/e7452d437493fc9345cd2e243d7fba960af71fdb/kinder/ci/tools/update-workflows/templates/workflows/skew-x-on-y-tasks.yaml#L68-L69

We will append`ignorePreflightErrors` after the `defaultIgnorePreflightErrors`.

kinder/ci/tools/update-workflows/templates/workflows/skew-kubelet-x-on-y.yaml
pacoxu: In this way, we have to wait to v1.32(probally) to remove all `ignorePreflightErrors: KubeletVersion` at that time.

kinder/ci/tools/update-workflows/templates/workflows/skew-x-on-y.yaml
pacoxu: We can remove `ignorePreflightErrors` suport and add `KubeletVersion` to the `defaultIgnorePreflightErrors` list then. And remove `KubeletVersion` from `defaultIgnorePreflightErrors` in v1.32.

kinder/ci/tools/update-workflows/templates/workflows/skew-kubelet-x-on-y.yaml
neolit123: > In this way, we have to wait to v1.32(probally) to remove all `ignorePreflightErrors: KubeletVersion` at that time.

SGTM

pacoxu:
I would suggest just changing the comments only in v1.29 if we want to keep it simple for maintenance.

I opened https://github.com/kubernetes/kubeadm/pull/2950.

kinder/ci/tools/update-workflows/templates/workflows/skew-x-on-y.yaml
neolit123: i see what the problem here is.
this file:
https://github.com/kubernetes/kubeadm/blob/e7452d437493fc9345cd2e243d7fba960af71fdb/kinder/ci/tools/update-workflows/templates/workflows/skew-kubelet-x-on-y.yaml

imports:
skew-x-on-y-tasks.yaml

so if you want to make changes to the kubelet skew it must be in skew-kubelet-x-on-y.yaml, not in skew-x-on-y-tasks.yaml. skew-x-on-y-tasks.yaml also applies to the kubeadm / control plane skew workflows.

if you add variables in skew-kubelet-x-on-y.yaml, they will act as overrides.

then again, this https://github.com/kubernetes/kubeadm/blob/e7452d437493fc9345cd2e243d7fba960af71fdb/kinder/ci/tools/update-workflows/templates/workflows/skew-kubelet-x-on-y.yaml#L10 comes from the config
https://github.com/kubernetes/kubeadm/blob/e7452d437493fc9345cd2e243d7fba960af71fdb/kinder/ci/tools/update-workflows/config.yaml#L95

so as long as all config entries have the KubeletVersion preflight error ingnored, it will work.

does this make sense to you?

neolit123 (commented):
@pacoxu 
test freeze is in two days.. you are probably busy with other things.
i will close this PR, send a new PR from me and ping you for review there.

pacoxu:
Thanks @neolit123

neolit123 (commented):
@pacoxu so after looking at this change and our comments from earlier again, we don't need to do changes to the tests. you already added the doc changes, those are ok.

we established that we can continue passing `--ignore-preflight-errors=KubeletVesion` even in cases where it's not needed.

that was because we don't want to update the config.yaml every release to target only specific versions with  `--ignore-preflight-errors=KubeletVesion`, this is annoying. instead we can just leave the test jobs they way they are and in 1 year when all supported kubeadm versions supported the new skew drop the  `--ignore-preflight-errors=KubeletVesion` for all of the jobs.

Issue #2414, opened 2021-03-19
Feature Reqest: Don't bind the admin.conf user to system:masters


raesene:
At the moment, when a new Kubeadm cluster is created, the default credential that is provided for cluster operators to use has a subject group of `system:masters`. e.g.

```
Subject: O = system:masters, CN = kubernetes-admin
```

This is a special group for which `cluster-admin` rights are hardcoded into the Kubernetes API server source code (https://github.com/kubernetes/kubernetes/blob/master/pkg/registry/rbac/escalation_check.go#L38).

As a result, this credential cannot be revoked, if a cluster operator loses control of it, or if one of their admins moves team or leaves the organization, the only way to revoke access completely is to rotate the cluster certificate authority keys.

Whilst Kubernetes in general doesn't currently support certificate revocation (https://github.com/kubernetes/kubernetes/issues/18982) it could be possible to reduce the risk posed by a lost admin.conf file, by having a specific clusterrolebinding for the admin.conf user to the cluster-admin clusterrole.

This would allow cluster operators to largely revoke the rights assigned to that credential, by removing the binding. It would still have `system:authenticated` rights, but these are likely (in most clusters) to be significantly lower than `cluster-admin`.

neolit123:
the admin.conf is like a root password...it must not be shared.
given i've seen a few cases where users share the file, i'm going to create a PR to make this very clear in https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm/.

it makes perfect sense for a deployer on cluster creation to sign break-glass credentials and that is what the admin.conf is.

> As a result, this credential cannot be revoked, if a cluster operator loses control of it, or if one of their admins moves team or leaves the organization, the only way to revoke access completely is to rotate the cluster certificate authority keys.

sharing the admin.conf in such a scenario is not ideal.
kubeadm gives you the `kubeadm kubeconfig user` command that can be used to create a kubeconfig for any user / group.

/kind documentation

raesene:
With the lifetime on the certificate being a year, it seems likely that in many cases users who had valid access to the credential when the cluster was created, may no longer need access during that period.

I'd agree that admin.conf should be a break-glass credential, but it may benefit cluster operators to provide a mechanism for that credential to be revoked, given its lifetime.

neolit123:
> With the lifetime on the certificate being a year, it seems likely that in many cases users who had valid access to the credential when the cluster was created, may no longer need access during that period.

[...]

PR #2951, opened 2023-11-01
kinder: fix missing quotes in super-admin-tasks.yaml

neolit123:
Without the quotes these (CMD) commands cannot be expanded properly.

xref https://github.com/kubernetes/kubeadm/pull/2949
xref https://github.com/kubernetes/kubeadm/issues/2414


pacoxu (commented):
/lgtm

PR #2950, opened 2023-11-01
Update skew change in v1.29 from n-1 to n-3

pacoxu:
- preflight check KubeletVersion behavior change

Only change the comments here.

Leave the remove of the KubeletVersion from ignorePreflightErrors to v1.32 when all skew CI can remove it.

PR #2953, opened 2023-11-03
Update skew change in v1.29 from n-1 to n-3

pacoxu:
- preflight check KubeletVersion behavior change

https://github.com/kubernetes/kubeadm/pull/2950


neolit123 (commented):
/lgtm

Issue #2910, opened 2023-08-02
1.29: housekeeping tasks


chendave:
This is a tracking issue for tasks that we are required to carry out during each cycle. **It tracks the upcoming release of Kubernetes 1.29 (1.x)**

Note: make sure you reference this issue in PRs.
 
same issue for 1.x-1 (see example PRs): #2857

 _When 1.x-4 version goes out of support_ https://kubernetes.io/releases/#release-history
 
* [x]   Drop kinder support for 1.x-4 and older (may not be needed every release)
[there are some TODOs for 1.25](https://github.com/search?q=repo%3Akubernetes%2Fkubeadm%201.25&type=code)\
- https://github.com/kubernetes/kubeadm/pull/2966

* [x]   Vendor the latest system-validators in kubernetes/kubernetes (may not be needed every release)
* [x]   Cleanup E2E jobs for 1.x-4 (requires two PRs)
  set `KUBERNETES_VERSION` to 1.x-1 and `KUBERNETES_SKEW` to `3` in /kinder/hack/update-workflows.sh
  make sure `PATH_TEST_INFRA` points to the right path.
  run the script and send two PRs for k/kubeadm and k/test-infra.
  See [KEP-3935: Generate test configurations for n-3 node skew #2855](https://github.com/kubernetes/kubeadm/pull/2855)

 _When the release-1.x branch and release-1.x test dashboards are created_
* [x]   Create E2E jobs for 1.x
  set `KUBERNETES_VERSION` to 1.x and `KUBERNETES_SKEW` to `3` in /kinder/hack/update-workflows.sh
  make sure `PATH_TEST_INFRA` points to the right path.
  run the script and send two PRs for k/kubeadm and k/test-infra.

- https://github.com/kubernetes/test-infra/pull/31194
- https://github.com/kubernetes/kubeadm/pull/2956

_Before the 1.x+1 cycle start_

* [ ]   Duplicate this issue for tracking 1.x+1 housekeeping tasks
* [ ]   Close this issue

neolit123:
/assign

the release-1.29 branch creation is scheduled for 1.29.

liangyuanpeng:
I' m come back after code freeze for 1.29 and still want to try it.
I cane see the below task need to work for 1.29:

- Update github action workflow for 1.29, same with https://github.com/kubernetes/kubeadm/pull/2855
- Update test-infra for KEP-3935, same with: https://github.com/kubernetes/test-infra/pull/29287
[...]

PR #2966, opened 2023-11-22
kinder: remove the handling of 1.25 and older versions of k8s

neolit123:
- Remove the repository replace from k8s.gcr.io -> registry.k8s.io
- Remove "kubelet-config-x.yy" handling, use "kubelet-config"
- Stop untainting "master" taints, only untaint control-plane taints
- Stop handling --experimental-patches, use --patches

xref https://github.com/kubernetes/kubeadm/issues/2910

### DIFF
--- a/kinder/pkg/cluster/manager/actions/kubeadm-init.go
+++ b/kinder/pkg/cluster/manager/actions/kubeadm-init.go
@@ -29,8 +29,6 @@ import (
 
 	"github.com/pkg/errors"
 
-	"k8s.io/apimachinery/pkg/util/version"
-
 	"k8s.io/kubeadm/kinder/pkg/cluster/manager/actions/assets"
 	"k8s.io/kubeadm/kinder/pkg/cluster/status"
 	"k8s.io/kubeadm/kinder/pkg/constants"
@@ -67,9 +65,9 @@ func KubeadmInit(c *status.Cluster, usePhases bool, copyCertsMode CopyCertsMode,
 
 	// execs the kubeadm init workflow
 	if usePhases {
-		err = kubeadmInitWithPhases(cp1, copyCertsMode, patchesDir, ignorePreflightErrors, vLevel)
+		err = kubeadmInitWithPhases(cp1, copyCertsMode, ignorePreflightErrors, vLevel)
 	} else {
-		err = kubeadmInit(cp1, copyCertsMode, patchesDir, ignorePreflightErrors, vLevel)
+		err = kubeadmInit(cp1, copyCertsMode, ignorePreflightErrors, vLevel)
 	}
 	if err != nil {
 		return err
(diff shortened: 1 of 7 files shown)


kinder/pkg/cluster/manager/actions/kubeadm-init.go
```diff
@@ -97,9 +95,7 @@ func kubeadmInit(cp1 *status.Node, copyCertsMode CopyCertsMode, patchesDir, igno
 		)
 	}
 	if patchesDir != "" {
-		if cp1.MustKubeadmVersion().LessThan(constants.V1_22) {
-			initArgs = append(initArgs, "--experimental-patches", constants.PatchesDir)
-		}
+		initArgs = append(initArgs, "--patches", constants.PatchesDir)
```
neolit123, line 98: --patches was completely missing in some places

pacoxu: IIRC, this is in the kubeadm config `/kind/kubeadm.conf` in new versions.
```
patches:
  directory: /kinder/patches
```

neolit123: right..thanks for the reminder.
only upgrade needs --patches

neolit123: updated!

pacoxu:
> will hold this until 1.29 is released.

/lgtm

kinder/pkg/cluster/manager/actions/kubeadm-init.go
```diff
@@ -229,13 +212,7 @@ func postInit(c *status.Cluster, wait time.Duration) error {
...
-		// TODO: Once kubeadm 1.23 is no longer supported remove the <1.24 handling.
-		// TODO: Remove only the "control-plane" taint for kubeadm >= 1.25.
-		// https://github.com/kubernetes/kubeadm/issues/2200
-		taints := []string{"node-role.kubernetes.io/control-plane-", "node-role.kubernetes.io/master-"}
-		if cp1.MustKubeadmVersion().LessThan(version.MustParseSemantic("v1.24.0")) {
-			taints = []string{"node-role.kubernetes.io/master-"}
-		}
+		taints := []string{"node-role.kubernetes.io/control-plane-"}
 		taintArgs := []string{"--kubeconfig=/etc/kubernetes/admin.conf", "taint", "nodes", "--all"}
 		taintArgs = append(taintArgs, taints...)
```
chendave, line 217: no longer to be a slice, fine as is.
```suggestion
		taint := "node-role.kubernetes.io/control-plane-"
		taintArgs := []string{"--kubeconfig=/etc/kubernetes/admin.conf", "taint", "nodes", "--all"}
		taintArgs = append(taintArgs, taint)
```

kinder/pkg/cluster/manager/actions/kubeadm-join.go
```diff
@@ -97,11 +97,6 @@ func kubeadmJoinControlPlane(cp *status.Node, patchesDir, ignorePreflightErrors
 		fmt.Sprintf("--ignore-preflight-errors=%s", ignorePreflightErrors),
 		fmt.Sprintf("--v=%d", vLevel),
 	}
```
chendave, line 99: I cannot comment on Line93, a quick question, does the parameter `patchesDir` is still needed, if it's only read from config file, I think we can remove it from func signature.

chendave (commented):
/lgtm

nice cleanup!

PR #2968, opened 2023-11-23
Update ha-considerations.md

adrienramiliharivelo:
frontend and backend name must be different

### DIFF
--- a/docs/ha-considerations.md
+++ b/docs/ha-considerations.md
@@ -143,12 +143,12 @@ frontend apiserver
     bind *:${APISERVER_DEST_PORT}
     mode tcp
     option tcplog
-    default_backend apiserver
+    default_backend apiserverbackend
 
 #---------------------------------------------------------------------
 # round robin balancing for apiserver
 #---------------------------------------------------------------------
-backend apiserver
+backend apiserverbackend
     option httpchk GET /healthz
     http-check expect status 200
     mode tcp


neolit123:
> frontend and backend name must be different

this is not true, according to the documentation.

see:
http://docs.haproxy.org/2.4/configuration.html

> A number of sections create an internal object or configuration space, which
requires to be distinguished from other ones. In this case they will take an
extra word which will set the name of this particular section. For some of them
the section name is mandatory. For example "frontend foo" will create a new
section of type "frontend" named "foo". Usually a name is specific to its
section and two sections of different types may use the same name, but this is
not recommended as it tends to complexify configuration management.

adrienramiliharivelo:
I produce the issue on this version of haproxy
HAProxy version 2.4.22-0ubuntu0.22.04.2 2023/08/14 - https://haproxy.org/
Status: long-term supported branch - will stop receiving fixes around Q2 2026.
Known bugs: http://www.haproxy.org/bugs/bugs-2.4.22.html
Running on: Linux 5.15.0-60-generic #66-Ubuntu SMP Fri Jan 20 14:29:49 UTC 2023 x86_64

The case is:
When I use the same name, the port of the load balancer is blocked.
But when I use different name, the port will opened.

neolit123 (commented):
maybe it's a bug in haproxy.

/cc @mbert
WDYT, seems fine if we add a unique name even if the docs say otherwise.

neolit123 (commented):
thanks @mbert !
/lgtm
/approve
/ok-to-test

PR #2969, opened 2023-11-23
kinder: optimize the preparation of upgrade artifacts

neolit123:
During the "altering" of the node image used by kinder, there are two download stages, one to download the init artifacts and one to download upgrade artifacts.

For some e2e workflows we have matching init and upgrade versions. This is done just to test the upgrade command and not perform an actual upgrade. It causes the same artifacts to be downloaded two times, which is slow.

During the preparation of bits installers, check if the init/upgrade versions match. If so, tell the upgrade bits installer to use the init directory as install source. This results in just copying the bits from the host, instead of downloading them again.

### DIFF
--- a/kinder/pkg/build/alter/alter.go
+++ b/kinder/pkg/build/alter/alter.go
@@ -146,6 +146,22 @@ func NewContext(options ...Option) (ctx *Context, err error) {
 
 // Alter alters the cluster node image
 func (c *Context) Alter() (err error) {
+	// create tempdir to alter the image in
+	alterDir, err := kindfs.TempDir("", "kinder-alter-image")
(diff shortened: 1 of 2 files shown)


neolit123:
it's actually possible to see this change in action on top of this presubmit job log for this same PR:
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubeadm/2969/pull-kubeadm-kinder-upgrade-latest/1727716099358724096/artifacts/task-01-add-kubernetes-versions-log.txt 

the artifacts for upgrade are just copied from init.

SataQiu (commented):
/lgtm

Issue #2971, opened 2023-11-24
kinder: include "upgrade plan" as part of upgrade workflows


neolit123:
"upgrade plan" is only covered by unit tests which arguably suffices, but ideally we should call it when we are doing upgrades in kinder too.

https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-upgrade/#cmd-upgrade-plan
we should include all flags that allow experimental / pre-release versions, also propagate preflight errors.

"upgrade plan" must be called on control plane nodes where "apply" is called.

see what places must be updated in kinder code and CI workflows;
https://github.com/search?q=repo%3Akubernetes%2Fkubeadm+%22apply%22+path%3Akinder&type=code

PR #2973, opened 2023-11-25
kinder: call "upgrade plan" before "upgrade apply"

neolit123:
During the kinder "kubeadm-upgrade" action, "upgrade apply"
is executed, but "upgrade plan" is not. To ensure better coverage
ensure "upgrade plan" is called as well. We have this command
as part of the kubeadm upgrade docs, so it makes sense to have
it here.

Ensure that "plan" is also called during the
"upgrade-latest-no-addon-config-maps" workflow.

NOTE: untested locally, but the presubmit job here should catch problems. i think this is all that has to be done.

fixes https://github.com/kubernetes/kubeadm/issues/2971

### DIFF
--- a/kinder/pkg/cluster/manager/actions/kubeadm-upgrade.go
+++ b/kinder/pkg/cluster/manager/actions/kubeadm-upgrade.go
@@ -53,6 +53,9 @@ func KubeadmUpgrade(c *status.Cluster, upgradeVersion *K8sVersion.Version, patch
 		}
 
 		if n.Name() == c.BootstrapControlPlane().Name() {
+			if err := kubeadmUpgradePlan(c, n, upgradeVersion, featureGate, vLevel); err != nil {
+				return err
+			}
 			err = kubeadmUpgradeApply(c, n, upgradeVersion, patchesDir, featureGate, wait, vLevel)
 		} else {
 			err = kubeadmUpgradeNode(c, n, upgradeVersion, patchesDir, wait, vLevel)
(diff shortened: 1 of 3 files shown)


neolit123:
/hold
the workflow upgrade-latest-no-addon-config-maps can have the explicit "upgrade plan" call too.

SataQiu (commented):
/lgtm

kinder/ci/tools/update-workflows/templates/workflows/dryrun-tasks.yaml
```diff
@@ -100,6 +100,7 @@ tasks:
   args:
     - -c
     - |
+      # TODO: "upgrade plan" does not have "--dry-run", yet. Add a "plan" call here, when it's supported.
```
chendave, line 103: IIRC, upgrade plan doesn't apply any change to the cluster, what  does the dry-run for the plan suppose to be?

kinder/pkg/cluster/manager/actions/kubeadm-upgrade.go
```diff
@@ -53,6 +53,9 @@ func KubeadmUpgrade(c *status.Cluster, upgradeVersion *K8sVersion.Version, patch
 		}
 
 		if n.Name() == c.BootstrapControlPlane().Name() {
+			if err := kubeadmUpgradePlan(c, n, upgradeVersion, featureGate, vLevel); err != nil {
```
chendave, line 56: we need to state it clearly that `kubeadm upgrade plan` can only run on the control node, maybe the `Short` msg in the cobra command, not related to this pr though.

chendave (commented):
/lgtm

kinder/ci/tools/update-workflows/templates/workflows/dryrun-tasks.yaml
neolit123: good point.

our dry run has a fake API client, so i think it means we can fake the calls that "plan" does as well?
or we could say that there is no plan to add --dry-run to "plan" completely.

kinder/pkg/cluster/manager/actions/kubeadm-upgrade.go
neolit123: it can be run on any node that has an admin.conf, i think.
but if you'd like to PR such a change on the cobra for "plan" please go ahead.

neolit123: handled ion separate PR
https://github.com/kubernetes/kubernetes/pull/122213

neolit123:
1.30 is out i'm going to unhold my kinder PRs one at a time to make sure we don't break many things at once.

neolit123:
@SataQiu @chendave 
ping, PTAL

SataQiu (commented):
/lgtm

PR #2979, opened 2023-12-05
kinder: remove parallel code during node creation

neolit123:
Running docker commands in parallel has proven to confuse the docker CR on the host and cause random failures. Also it is less clear what container command failed exactly.

Make the whole node creation linear. This will inevitably make the kinder cluster creation slower, but will reduce the change for unexplainable flakes and allow us to debug and potentially retry certain commands that are flaky.

In the past one such command was a simple:
    "docker exec ... mkdir ..."

### DIFF
--- a/kinder/pkg/cluster/manager/create.go
+++ b/kinder/pkg/cluster/manager/create.go
@@ -148,7 +148,7 @@ func CreateCluster(clusterName string, options ...CreateOption) error {
 		clusterName,
 		flags,
 	); err != nil {
-		return handleErr(err)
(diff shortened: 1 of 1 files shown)


neolit123:
> This will inevitably make the kinder cluster creation slower, but will reduce the change for unexplainable flakes and allow us to debug and potentially retry certain commands that are flaky.

looking at logs it still takes around 2 minutes for 6 nodes cluster to be created in prow CI

chendave (commented):
/lgtm

PR #2983, opened 2023-12-21
Fixed Concurrency Issue: channel communication of kinder

fazledyn-or:
While triaging your project, our bug fixing tool generated the following message-

> In File: [waiter.go](https://github.com/kubernetes/kubeadm/blob/main/kinder/pkg/cluster/manager/actions/waiter.go#L154), Line: 154: `pass <- true`. In function `waitFor`, there is a send operation. But no matching receive operation is found on that channel from any other goroutine.

## Description
After manual triaging, we found out that the go routine inside the `waitFor` method is sending to the `pass` channel. And the polling loop inside the `waitFor` method is receiving from the channel. We also noted that there's a `timer` inside the `select` block that returns False if the timeout occurs.

It all seems correct, except a corner case such as-

The receive from `pass` and the `timer.C` occurs at the same time. In that case, the `select` will randomly choose one of the channels. From [documentation](https://go.dev/ref/spec#Select_statements)-

> If one or more of the communications can proceed, a single one that can proceed is chosen via a uniform pseudo-random selection. Otherwise, if there is a default case, that case is chosen. If there is no default case, the "select" statement blocks until at least one of the communications can proceed.

In that case, if the `timer.C` is chosen, the go routine which has already sent `True` to `pass` will remain halted, expecting another go routine to receive.

## Solution
To address this case, we can close the `pass` channel inside the `timer.C` case, which will notify the blocked go routines. In order to do so, we must reverse the channel operation. Because, sending data on a closed channel will throw a `panic` whereas receiving won't.

## CLA Requirements
*This section is only relevant if your project requires contributors to sign a Contributor License Agreement (CLA) for external contributions.*

All contributed commits are already automatically signed off.

> The meaning of a signoff depends on the project, but it typically certifies that committer has the rights to submit this work under the same license and agrees to a Developer Certificate of Origin (see [https://developercertificate.org/](https://developercertificate.org/) for more information).
>
> \- [Git Commit SignOff documentation](https://developercertificate.org/)

## Sponsorship and Support
This work is done by the security researchers from OpenRefactory and is supported by the [Open Source Security Foundation (OpenSSF)](https://openssf.org/): [Project Alpha-Omega](https://alpha-omega.dev/). Alpha-Omega is a project partnering with open source software project maintainers to systematically find new, as-yet-undiscovered vulnerabilities in open source code - and get them fixed – to improve global software supply chain security.

The bug is found by running the Intelligent Code Repair (iCR) tool by OpenRefactory and then manually triaging the results.

### DIFF
--- a/kinder/pkg/cluster/manager/actions/waiter.go
+++ b/kinder/pkg/cluster/manager/actions/waiter.go
@@ -151,7 +151,7 @@ func waitFor(c *status.Cluster, n *status.Node, timeout time.Duration, condition
 
 			for {
 				if x(c, n) {
-					pass <- true
+					<-pass
 					break
 				}
 				// add a little delay + jitter before retry
@@ -164,12 +164,14 @@ func waitFor(c *status.Cluster, n *status.Node, timeout time.Duration, condition
 	passed := 0
 	for {
 		select {
-		case <-pass:
+		case pass <- true:
 			passed++
 			if passed == len(conditions) {
 				return true
 			}
 		case <-timer.C:
+			// close the channel if timeout occurs, this will release all the blocked receives
+			close(pass)
 			return false
 		}
 	}


neolit123 (commented):
thanks for the PR and detailed description

> The receive from pass and the timer.C occurs at the same time. In that case, the select will randomly choose one of the channels. From [documentation](https://go.dev/ref/spec#Select_statements)-

that's not likely to happen, but the change seems correct to me.
let's see if our CI is happy with it.

/ok-to-test

neolit123 (commented):
the function watiFor() sees its usage in the same file and only a couple of functions benefit from the parallelism. kinder is not so dependent on performance, so we can actually remove waitFor() and refactor the dependent waiters at some point.

neolit123:
please see the CI error:

```
[*] Verifying gofmt...
diff pkg/cluster/manager/actions/waiter.go.orig pkg/cluster/manager/actions/waiter.go
--- pkg/cluster/manager/actions/waiter.go.orig
+++ pkg/cluster/manager/actions/waiter.go
@@ -164,7 +164,7 @@
 	passed := 0
 	for {
 		select {
-		case pass<-true:
+		case pass <- true:
 			passed++
 			if passed == len(conditions) {
 				return true

```

fazledyn-or:
> please see the CI error:
> 

Formatted the code. Should be okay now.

neolit123:
> > please see the CI error:
> 
> Formatted the code. Should be okay now.

please squash the commits to 1.

neolit123:
/cc @chendave 
for additional review, PTAL.

fazledyn-or:
> > > please see the CI error:
> > 
> > 
> > Formatted the code. Should be okay now.
> 
> please squash the commits to 1.

Done. Please have a look.

neolit123 (commented):
/approve
needs a 2nd reviewer

NOTE: the swap of send/receive works because the channel is unbuffered, see:
https://stackoverflow.com/a/47525746

>> If the channel is unbuffered, the sender blocks until the receiver has received the value. If the channel has a buffer, the sender blocks only until the value has been copied to the buffer; if the buffer is full, this means waiting until some receiver has retrieved a value.

neolit123 (commented):
/lgtm
/approve

tested locally. should be good to go.

PR #3012, opened 2024-02-06
kinder: update super-admin workflow for 1.30

neolit123:
xref https://github.com/kubernetes/kubeadm/issues/2414#issuecomment-1929099638

kubeadm 1.29 had code that creates the super-admin.conf file on upgrade. the same code is removed in kubeadm 1.30. Update the super-admin kinder workflow to reflect this change.

PR #3015, opened 2024-02-07
kinder: apply yet another fix for super-admin workflow

neolit123:
The super-admin.conf file was no longer expected after upgrade, thus we don't have to read the file and check its subject.

follow up on https://github.com/kubernetes/kubeadm/pull/3012#issuecomment-1931579134


pacoxu (approved):
/lgtm

PR #3035, opened 2024-03-10
Remove --enable-hostpath-provisioner option

carlory:
See https://github.com/kubernetes/kubernetes/issues/123804#issuecomment-1985428951

/cc @neolit123


neolit123 (commented):
/lgtm
/approve
thanks

cc @pacoxu

PR #3037, opened 2024-03-19
test: add support for OCI images

rchincha:
Partially fixes issue
https://github.com/kubernetes/kubernetes/issues/121742


neolit123 (commented):
what is the status of this @rchincha ?

i don't see a KEP for the overall change.
https://github.com/kubernetes/kubernetes/issues/121742#issuecomment-1794892811

also you don't have the CLA signed
so we cannot accept PRs from you https://github.com/kubernetes/kubeadm/pull/3037#issuecomment-2008219885

PR #3056, opened 2024-05-03
kinder: include upgrades and cluster-info in external CA workflows

neolit123:
The workflow already includes upgrade artifacts with the same k8s version. Perform upgrade and do cluster-info and kubeadm-e2e before and after the upgrade.

xref https://github.com/kubernetes/kubeadm/issues/3055

/hold
hold until https://github.com/kubernetes/kubernetes/pull/124682 merges as these changes will fail on 1.29 otherwise.


SataQiu (commented):
/lgtm

PR #3057, opened 2024-05-03
kinder: prepull the kindnet image when building the node image

neolit123:
The CNI (kindnet) image was the only image that was not prepulled during the creation of a node image. This allows kinder to deploy the cluster without every node having to pull the CNI image.


SataQiu (commented):
/lgtm

PR #3058, opened 2024-05-04
Update pull.go: Fix Typo, And a suggestion

1000oaks:
I was wondering why each time I use `kinder create cluster ...`, connections to `auth.docker.io` and `registry-1.docker.io` would be made. Then I found this typo.

Besides, to keep in accordance with KinD, I think that `imagePullPolicy: "Never"` can be provided **(or becoming configurable)** in `kinder/pkg/kubeadm/config.go` (variable `configTemplateBetaV2` and `configTemplateBetaV3`) to make clustion creation process fully offline-friendly.

If I were wrong, please point out. Thanks!

edit neolit123
fixes https://github.com/kubernetes/kubeadm/issues/3059

### DIFF
--- a/kinder/pkg/cri/host/pull.go
+++ b/kinder/pkg/cri/host/pull.go
@@ -28,7 +28,7 @@ import (
 func PullImage(image string, retries int) (bool, error) {
 	// once we have configurable log levels
 	// if this did not return an error, then the image exists locally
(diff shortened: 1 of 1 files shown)


neolit123 (commented):
change seems ok, but you need to sign the CLA

neolit123 (commented):
/lgtm
/approve
/ok-to-test

thanks

PR #3070, opened 2024-06-01
kinder: support patching OCI image tar archives

neolit123:
When kinder is building a node image it fetches image TAR files for the k8s components from the CI bucket. These image TARs contain spec tagged as '-amd64", so they are patched to not have this suffix before
being imported in containerd. This makes them compatible with kubeadm commands.

Recently the image format changed from Docker to OCI. To support OCI, parse an additional "index.json" file which is part of the OCI image spec.

Docker image TARs continue to be supported.

----

notified SIG release here:
https://kubernetes.slack.com/archives/C2C40FMNF/p1717258237762459

fixes failing kubeadm CI:
https://testgrid.k8s.io/sig-cluster-lifecycle-kubeadm
https://gcsweb.k8s.io/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-kubeadm-kinder-learner-mode-latest/1796596050144792576/

during `init` kinder reports that "-amd64" images are preloaded:
```
time="17:38:01" level=debug msg="Running: docker exec kinder-learner-mode-control-plane-1 ctr --namespace=k8s.io images ls -q"
time="17:38:01" level=debug msg="List of images already pre-loaded in the node [docker.io/kindest/kindnetd:0.5.4 registry.k8s.io/coredns/coredns:v1.11.1 registry.k8s.io/etcd:3.5.13-0 registry.k8s.io/kube-apiserver-amd64:v1.31.0-alpha.0.1047_2f0cf23645876d 
...
```

`init` fails with apiserver error:
```
I0531 17:42:10.288428     291 round_trippers.go:553] GET https://172.17.0.7:6443/healthz?timeout=10s  in 0 milliseconds
I0531 17:42:11.288884     291 with_retry.go:234] Got a Retry-After 1s response for attempt 9 to https://172.17.0.7:6443/healthz?timeout=10s
I0531 17:42:11.289681     291 round_trippers.go:553] GET https://172.17.0.7:6443/healthz?timeout=10s  in 0 milliseconds
[api-check] The API server is not healthy after 4m0.351859438s

Unfortunately, an error has occurred:
	context deadline exceeded
```

kubelet.log (image not imported with expected name)
```
\"registry.k8s.io/kube-apiserver:v1.31.0-alpha.0.1047_2f0cf23645876d\": failed to resolve reference \"registry.k8s.io/kube-apiserver:v1.31.0-alpha.0.1047_2f0cf23645876d\": registry.k8s.io/kube-apiserver:v1.31.0-alpha.0.1047_2f0cf23645876d: not found" image="registry.k8s.io/kube-apiserver:v1.31.0-alpha.0.1047_2f0cf23645876d"
```

### DIFF
--- a/kinder/pkg/cri/host/archive.go
+++ b/kinder/pkg/cri/host/archive.go
@@ -24,6 +24,8 @@ import (
 	"io"
 	"os"
 	"strings"
+
+	log "github.com/sirupsen/logrus"
 )
 
 // TODO: investigate if we can get rid of this or simplify it.
@@ -79,14 +81,17 @@ func GetArchiveTags(path string) ([]string, error) {
 
 // EditArchiveRepositories applies edit to reader's image repositories,
 // IE the repository part of repository:tag in image tags
-// This supports v1 / v1.1 / v1.2 Docker Image Archives
 //
-// editRepositories should be a function that returns the input or an edited
-// form, where the input is the image repository
-//
+// It supports v1 / v1.1 / v1.2 Docker Image Archives:
 // https://github.com/moby/moby/blob/master/image/spec/v1.md
 // https://github.com/moby/moby/blob/master/image/spec/v1.1.md
 // https://github.com/moby/moby/blob/master/image/spec/v1.2.md
+//
+// and OCI schema v2 with index.json:
+// https://github.com/opencontainers/image-spec/blob/main/spec.md
+//
+// editRepositories should be a function that returns the input or an edited
+// form, where the input is the image repository
 func EditArchiveRepositories(reader io.Reader, writer io.Writer, editRepositories func(string) string) error {
 	tarReader := tar.NewReader(reader)
 	tarWriter := tar.NewWriter(writer)
@@ -104,15 +109,25 @@ func EditArchiveRepositories(reader io.Reader, writer io.Writer, editRepositorie
 			return err
 		}
 
-		// edit the repostories and manifests files when we find them
+		// A Docker image archive has a 'repositories' and a 'manifest.json' files.
+		// An OCI image archive has the same 'manifests.json' file format, but also an 'index.json' file.
 		if hdr.Name == "repositories" {
+			log.Infof("editing %s", hdr.Name)
 			b, err = editRepositoriesFile(b, editRepositories)
 			if err != nil {
 				return err
 			}
 			hdr.Size = int64(len(b))
 		} else if hdr.Name == "manifest.json" {
+			log.Infof("editing %s", hdr.Name)
 			b, err = editManifestRepositories(b, editRepositories)
+			if err != nil {
+				return err
+			}
+			hdr.Size = int64(len(b))
+		} else if hdr.Name == "index.json" {
+			log.Infof("editing %s", hdr.Name)
+			b, err = editIndexRepositories(b, editRepositories)
 			if err != nil {
 				return err
 			}
@@ -161,7 +176,6 @@ type metadataEntry struct {
 	Layers   []string `json:"Layers"`
 }
 
-// applies
 func editManifestRepositories(raw []byte, editRepositories func(string) string) ([]byte, error) {
 	var entries []metadataEntry
 	if err := json.Unmarshal(raw, &entries); err != nil {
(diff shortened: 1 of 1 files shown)


kinder/pkg/cri/host/archive.go
```diff
@@ -185,6 +199,44 @@ func editManifestRepositories(raw []byte, editRepositories func(string) string)
 	return json.Marshal(entries)
 }
 
+// https://github.com/opencontainers/image-spec/blob/main/image-index.md
+type index struct {
```
mikebrow, line 203: Editing contents of oci objects in this way seems problematic. https://github.com/opencontainers/image-spec/blob/v1.0/specs-go/v1/index.go#L21 https://github.com/opencontainers/image-spec/blob/v1.0/specs-go/v1/descriptor.go#L55 These types have major changes in oci 1.1 and even 1.0 seems different. Maybe I need more context, on my phone ATM.

mikebrow (commented):
Suggest considering using version correct oci spec structures.

kinder/pkg/cri/host/archive.go
neolit123: the thing here is a hack that has to be babysitted anyway if k8s switches to non backwards compatible spec that kinder does not support.

as long as k8s image maintainers tell us  all should be good.

neolit123:
CI got green again without this PR merging. i think it was this revert that fixed it:
https://github.com/kubernetes/kubernetes/pull/125269

either way, i think this can be merged as a backup scenario e.g. so that we don't have to debug on a weekend someday...
if k8s has smoother plans to migrate the images to a new spec this code can be better organized.

kinder/pkg/cri/host/archive.go
mikebrow: Which spec does kinder support? If you wanted to support the 3 valid spec schemas, docker 2, oci 1.0 and 1.1 you could switch on type before doing the conversions. But I'm still not sure why these files are being edited.

neolit123: with this PR oci v2, before it only docker 1.0, 1.1, 1.2.

the k8s image tars from CI are amd-64 only. kinder here is making edits to change them from "-amd64" to "", so that kubeadm can work with these images. kubeadm needs images that are multi-os-arch (manifest lists) without arches/os in the image names.

kinder/pkg/cri/host/archive.go
```diff
@@ -185,6 +199,44 @@ func editManifestRepositories(raw []byte, editRepositories func(string) string)
...
+}
+
+func editIndexRepositories(raw []byte, editRepositories func(string) string) ([]byte, error) {
+	var idx index
+	if err := json.Unmarshal(raw, &idx); err != nil {
+		return nil, err
+	}
+
+	for _, m := range idx.Manifests {
+		const containerdImage = "io.containerd.image.name"
```
carlory, line 224: The annotation `io.containerd.image.name` is not a part of the standard oci image spec. All pre-Defined Annotation Keys can be found [here](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys). There is no `io.containerd.image.name` in the list. but oci image spec allows for arbitrary annotations. so the annotation is added by containerd (may be buildkit).

`org.opencontainers.image.ref.name` is the pre-defined annotation for name of the reference for a target. 

In containerd, `io.containerd.image.name` stores the fully qualified reference name of a image. it includes the registry, repository, tag. but `org.opencontainers.image.ref.name` stores only the tag of the image.

```
{
  "annotations": {
    "io.containerd.image.name": "docker.io/library/alpine:latest",
    "org.opencontainers.image.ref.name": "latest"
  }
}
```

I'm not sure whether the `org.opencontainers.image.ref.name` can be used to store a fully qualified reference name of a image. The image spec does not say anything about it. 

```
{
  "annotations": {
    "org.opencontainers.image.ref.name": "docker.io/library/alpine:latest"
  }
}
```

But it is possible to store the fully qualified reference name of a image because the validation of the annotation allows this. The rules for the annotation value are:

```
ref       ::= component ("/" component)*
component ::= alphanum (separator alphanum)*
alphanum  ::= [A-Za-z0-9]+
separator ::= [-._:@+] | "--"
```

As the OCI spec was based on the Docker image format, the containerd may more closely follow the oci image spec. However, in my mind, I'm not sure whether there is a certain reason for containerd to use `io.containerd.image.name` instead of `org.opencontainers.image.ref.name`, maybe it is just a historical reason or a design decision.

In kinder, only getting the image name from the annotation `io.containerd.image.name` makes sense.

Just for the record, the containerd gets the image name from annotaions in the following order:

```
1. io.containerd.image.name if present
2. org.opencontainers.image.ref.name

code: https://github.com/containerd/containerd/blob/main/client/import.go#L257
```

cc @neolit123

neolit123: this is what i tested and it worked. the annotation was in the yaml and if it was patched the image was imported with a different name with ctr, which is what we want.

the bigger question is why it happened. i got some indication on slack it happened because the docker version used to produce the tars was updated in k/release. but then it was reverted. the people who did all that did not reply.

this pr is currently not needed and can be closed. hopefully sig release will write an email about such spec changing in the future on the k-dev mailing list.

PR #3120, opened 2024-11-02
kinder: add tasks file for all workflows

neolit123:
This makes it consistent with the rest of the workflow files that are all pairs:
- main workflow with version
- tasks file for the workflow


pacoxu (commented):
/lgtm

PR #3123, opened 2024-11-13
kubeadm: increase timeout for dryrun-without-cluster

SataQiu:
kubeadm: increase timeout for dryrun-without-cluster

Hopefully, the execution can be complete in 15 minutes. Not sure what timeout should be set to.

fix: https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/ci-kubernetes-e2e-kubeadm-kinder-dryrun-latest/1856721175695069184


neolit123 (commented):
hm, the whole dry run should probably take a couple of minutes on a good machine.
maybe that particular run in prow was slow.

we could wait for another one before increasing the number?

neolit123:
/lgtm
/hold

you can cancel the hold if you think the increase is needed.

pacoxu:
/test pull-kubeadm-kinder-upgrade-latest
https://testgrid.k8s.io/sig-cluster-lifecycle-kubeadm#kubeadm-kinder-dryrun-latest keeps timeout

pacoxu:
Does the presubmit yaml need an update as well?
https://github.com/kubernetes/kubeadm/blob/main/kinder/ci/workflows/presubmit-upgrade-tasks.yaml

Issue #3042, opened 2024-04-05
FEATURE REQUEST: Add kubelet instance config


HirazawaUi:
## Is this a BUG REPORT or FEATURE REQUEST?

FEATURE REQUEST

## Anything else we need to know?

This issue wants to remove container runtime interface (CRI) socket annotations from Node objects in Kubernetes and no longer added `--container-runtime-endpoint` args in `kubeadm-flags.env` ( `--container-runtime-endpoint` args deprecated in kubelet), this annotation and args are used to specify the CRI socket endpoint on each node that the kubelet uses to communicate with the container runtime.

Add `kubelet-instance-config.yml` locally to store container runtime interface (CRI) sockets, and in `kubeadm init`, `join` and `upgrade`, override the global kubelet configuration through `kubelet-instance-config.yml`.

More detailed design at: https://hackmd.io/@r2Yq-PVoRcK2OaGypnwPMg/rJFChACCp/edit

neolit123:
@HirazawaUi we need a KEP for this, ket's see what @pacoxu can comment about the previous work done on this.

@pacoxu can you please post links to your previous KEP work.?
also didn't we already have a tracking issue here in k/kubeadm?

maybe we can close KEP PRs and let @HirazawaUi to take over.

HirazawaUi:
> also are you willing to work on this @HirazawaUi there are a lot of things to update around this change across multiple releases. and i don't think me or anyone else will have a lot of time for it.
> 
> * KEP design how is it going to work exactly, do we need a feature gate
> * k8s.io docs update
> * k/k code update
> * do we need a e2e test in k/kubeadm
> 
> too many questions

Yes, I'll try to finish it.

HirazawaUi:
@pacoxu What do you think about this? I would love to hear your comments :)

pacoxu:
The KEP 
- https://github.com/kubernetes/enhancements/issues/3929
- https://github.com/kubernetes/enhancements/pull/3930

My current proposal would be something like  https://github.com/kubernetes/enhancements/pull/3930#issuecomment-1931657825
> > i meant a generic patches approach where a local config file overrides the global config downloaded from the cm. not the --patches feature.
> > 
> > this local file must be stored somewhere. perhaps in the same dir as config.yaml, but called config-instance.yaml. IMO it has a number of tricky aspects that need to be covered in the design doc for init, join, upgrade.
> > 

[...]

PR #3126, opened 2024-11-18
Add e2e test for kubelet instance config

HirazawaUi:
Add e2e test for kubelet instance config.

relate: https://github.com/kubernetes/test-infra/pull/33807

--- edit neolit123
xref
- https://github.com/kubernetes/kubeadm/issues/3042


kinder/ci/tools/update-workflows/templates/testinfra/kubeadm-kinder-instance-config.yaml
```diff
@@ -0,0 +1,42 @@
...
+  interval: {{ .JobInterval }}
+  decorate: true
+  labels:
+    preset-dind-enabled: "true"
+    preset-kind-volume-mounts: "true"
+  annotations:
+    testgrid-dashboards: sig-cluster-lifecycle-kubeadm
+    testgrid-tab-name: kubeadm-kinder-instance-config-{{ dashVer .KubernetesVersion }}
+    testgrid-alert-email: sig-cluster-lifecycle-kubeadm-alerts@kubernetes.io
+    description: "OWNER: sig-cluster-lifecycle (kinder); Uses kubeadm/kinder to create a cluster using kubelet instance config and run kubeadm-e2e and the conformance suite"
```
neolit123, line 12: i think we can skip the conformance suite, because it overlaps with other jobs.

kinder/ci/tools/update-workflows/templates/workflows/instance-config-tasks.yaml
```diff
@@ -0,0 +1,185 @@
...
+    Runs kubeadm e2e tests
+  cmd: kinder
+  args:
+  - test
+  - e2e-kubeadm
+  - --test-flags=--report-dir={{ .env.ARTIFACTS }} --report-prefix=e2e-kubeadm
+  - --name={{ .vars.clusterName }}
+  - --loglevel=debug
+  timeout: 10m
+- name: e2e
```
neolit123, line 136: we can remove the conformance e2e task

neolit123 (commented):
i think we should do the following:

- kubeadm init with "latest"
- kubeadm join with "latest"
- check that that instance config is on the nodes
- delete the instance config from the nodes
- kubeadm upgrade "latest" (it would be latest->latest techniclly)
- check if the instance config is on the nodes

kinder/ci/tools/update-workflows/templates/workflows/instance-config-tasks.yaml
```diff
@@ -0,0 +1,214 @@
...
+  args:
+  - do
+  - kubeadm-upgrade
+  - --upgrade-version={{ .vars.upgradeVersion }}
+  - --name={{ .vars.clusterName }}
+  - --loglevel=debug
+  - --kubeadm-verbosity={{ .vars.kubeadmVerbosity }}
+  - --kubeadm-feature-gate="NodeLocalCRISocket=true"
+  timeout: 15m
+- name: check-config
```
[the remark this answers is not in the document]
HirazawaUi: fixed.

neolit123 (commented):
/lgtm
/approve
thanks

kinder/ci/workflows/instance-config-tasks.yaml
```diff
@@ -0,0 +1,214 @@
...
+- name: remove-config
+  description: |
+    Remove the kubelet instance configuration from work node.
+  cmd: /bin/bash
+  args:
+    - -c
+    - |
+      set -x
+
+      for node in worker-{1..2}; do
```
HirazawaUi, line 112: We only need to delete the instance kubelet configuration for worker nodes. For control-plane nodes, they must obtain the criSocket from either the annotation or the local instance kubelet configuration.

This behavior is expected: for control-plane nodes, the instance kubelet configuration file is generated only when kubeadm init is executed with the NodeLocalCRISocket feature gate enabled, and the kubeadm.alpha.kubernetes.io/cri-socket annotation will not added to the node object.

In this case, deleting the instance kubelet configuration file will result in an error state.

HirazawaUi: Yes, both control plane nodes and worker nodes will write to the kubelet instance configuration. However, during the execution of `kubeadm upgrade node`, the control plane nodes will retrieve the CRI information either from the Node object or the kubelet instance configuration.

If the feature gate was enabled when the cluster was initialized using kubeadm init, the Node objects will not have the `kubeadm.alpha.kubernetes.io/cri-socket` annotation. If the kubelet instance configuration is deleted, an error will occur during `kubeadm upgrade` because the CRI information cannot be retrieved.

HirazawaUi: > 3\. upgrading from FG disabled -> FG enabled
> 
> * delete instance config
> * apply the node annotations with kubectl
> * call upgrade
> * make sure instance-config is present

It seems there is an issue. The instance config did not exist before the upgrade.

The correct process should be:

* create the cluster with the feature gate disabled.
* verify that the annotation exists and the instance config does not exist.
* call the upgrade with the feature gate enabled.
* verify that the annotation not exists and the instance config is present.

neolit123: we need to create the cluster with FG enabled to verify instance config will be created.
note both 1 and 2 are steps in the same kinder workflow file.

then my steps in 2 make sure that if the FG is enabled (in clusterconfiguration) but if the instance config is missing it would use the annotation or flags on disk values to migrate the user to start using the FG and instance config. is that not a supported case in the code in k/k?

kubeadm feature gates can also be enabled before upgrade. by the user manually editing the kubeadm-config CM.

HirazawaUi: > we need to create the cluster with FG enabled to verify instance config will be created.
> note both 1 and 2 are steps in the same kinder workflow file.

Ah, sorry, I initially thought it needed two files, like this:
```
tasks:
- import: instance-config-enabled-upgrade-to-enabled-tasks.yaml
- import: instance-config-disabled-upgrade-to-enabled-tasks.yaml
```

I've got what you mean. Thank you for your patient explanation.

pacoxu:
BTW, @neolit123 is it neccessary for us to add some CIs like https://testgrid.k8s.io/sig-release-master-blocking#gce-cos-master-alpha-features to enable all alpha or beta features during the upgrade and init? This is out of the scope of this PR.

kinder/ci/workflows/instance-config-tasks.yaml
```diff
@@ -0,0 +1,214 @@
...
+  description: |
+    Join the other nodes to the Kubernetes cluster
+  cmd: kinder
+  args:
+  - do
+  - kubeadm-join
+  - --name={{ .vars.clusterName }}
+  - --loglevel=debug
+  - --kubeadm-verbosity={{ .vars.kubeadmVerbosity }}
+  - --kubeadm-feature-gate="NodeLocalCRISocket=true"
```
pacoxu, line 76: only init with the FG.
- we can enable the FG before join and upgrade.

https://github.com/kubernetes/kubeadm/blob/33b7812d98b7f202fdae20a4810e7dfbf033da33/kinder/pkg/cluster/manager/actions/actions.go#L40-L53

neolit123:
> BTW, @neolit123 is it neccessary for us to add some CIs like https://testgrid.k8s.io/sig-release-master-blocking#gce-cos-master-alpha-features to enable all alpha or beta features during the upgrade and init? This is out of the scope of this PR.

i personally think that our current setup is better which is - one dedicated e2e for each FG.
if we have FGs that interact with each other we might want to allow kinder to enable  >1 but that's more of a corner case.

kinder/ci/workflows/instance-config-tasks.yaml
neolit123: only calling the FG on init is fine, because it will be stored in the CM.
this plan will work https://github.com/kubernetes/kubeadm/pull/3126#discussion_r1866071594

kinder/ci/tools/update-workflows/templates/workflows/instance-config-tasks.yaml
```diff
@@ -0,0 +1,340 @@
...
+  - kubeadm-join
+  - --name={{ .vars.clusterName }}
+  - --loglevel=debug
+  - --kubeadm-verbosity={{ .vars.kubeadmVerbosity }}
+  - --kubeadm-feature-gate="NodeLocalCRISocket=true"
+  - --copy-certs=auto
+  timeout: 10m
+- name: check-annotation
+  description: |
+    check the cri-socket annotation not exists on each node.
```
neolit123, line 80:
```suggestion
    Check that the cri-socket annotation does not exist on each node.
```

kinder/ci/tools/update-workflows/templates/workflows/instance-config-tasks.yaml
```diff
@@ -0,0 +1,340 @@
...
+        CMD="docker exec {{ .vars.clusterName }}-$node"
+
+        ${CMD} bash -c 'rm -f /var/lib/kubelet/instance-config.yaml'
+      done
+
+      exit 0
+  timeout: 5m
+- name: add-cri-socket-annotation
+  description: |
+    Add annotation to each node object
```
neolit123, line 184:
```suggestion
    Add the cri-socket annotation to each node.
```

kinder/ci/tools/update-workflows/templates/workflows/instance-config-tasks.yaml
```diff
@@ -0,0 +1,340 @@
...
+          else
+            echo "Failed to add annotation to node $node."
+            exit 1
+          fi
+        done
+      '
+
+      exit 0
+  timeout: 5m
+- name: disable-features-gate
```
neolit123, line 210: we shouldn't disable the FG.

if the FG is enabled before upgrade manually by the user it should transform the cluster.
- removing the annotation from each node object (maybe we can do this for the beta)
- writing the instance-config.yaml on each node
-

HirazawaUi: > we shouldn't disable the FG.

I think whether the feature gate was enabled before the upgrade is not important to us.

For the new functionality we introduced, we won’t check whether the feature gate was enabled previously; we will only verify its annotation and instance-config.yaml.

Here, we simulate a cluster that has never used the `NodeLocalCRISocket` feature gate before. It has the `kubeadm.alpha.kubernetes.io/cri-socket` annotation and no `instance-config.yaml`. During the upgrade, we enable the feature gate and observe whether the subsequent behavior meets our expectations.

HirazawaUi: Or I have misunderstood your meaning.

neolit123: here is the test that we need explained again. some of these might need k/k code changes.

- keep the FG enabled in the cluster
- make sure the annotation is present
- make sure that the instance config is not present
- run upgrade
- make sure that instance config was generated
- make sure that the --container-runtime-endpoint flag is removed from the kubeadm-flags.env (needs k/k change)
- make sure that the annotation was removed (needs k/k change)

this covers the scenario where users will enable the FG _before_ upgrading.

HirazawaUi: > here is the test that we need explained again. some of these might need k/k code changes.
> 
> * keep the FG enabled in the cluster
> * make sure the annotation is present
> * make sure that the instance config is not present
> * run upgrade
> * make sure that instance config was generated
> * make sure that the --container-runtime-endpoint flag is removed from the kubeadm-flags.env (needs k/k change)
> * make sure that the annotation was removed (needs k/k change)

I now understand the core of our disagreement. I should not disable the `NodeLocalCRISocket` feature gate in `kubeadm-config` before the kubeadm upgrade and then attempt to enable it during `kubeadm upgrade` using the `--feature-gates` parameter.

This is because `kubeadm upgrade` does not accept the `--feature-gates` parameter. For it, the only way to determine whether a feature gate is enabled is through the `feature-gates` defined in `kubeadm-config`.

I think the current test tasks already cover the relevant cases:

* task-add-cri-socket-annotation: Ensure the annotation is present.
* task-remove-config: Ensure the instance config is not present.
* task-upgrade-again: Run the upgrade process.
* task-check-config-after-second-upgrade: Ensure the instance config was generated.

But...., there is another issue to address. If we disable the `NodeLocalCRISocket` feature gate in `kubeadm-config`, then the `task-check-config-after-second-upgrade` should fail. This is because the `--kubeadm-feature-gates="NodeLocalCRISocket=true"` parameter added in `task-upgrade-again` is ineffective. I will investigate tomorrow why this is happening.

HirazawaUi: > But...., there is another issue to address. If we disable the `NodeLocalCRISocket` feature gate in `kubeadm-config`, then the `task-check-config-after-second-upgrade` should fail. This is because the `--kubeadm-feature-gates="NodeLocalCRISocket=true"` parameter added in `task-upgrade-again` is ineffective. I will investigate tomorrow why this is happening.

The reason is that the scripts in `check-config-after-second-upgrade` and `check-config` always end with exit code 0 because I did not capture the return result of the commands executed inside the container, and the script always ends with exit 0.


I have removed the disable-features-gate task and modified all the scripts to capture the return values from the container. Please review it again to see if it now meets our expectations.

kinder/ci/tools/update-workflows/templates/workflows/instance-config-tasks.yaml
```diff
@@ -0,0 +1,322 @@
...
+  description: |
+    Join the other nodes to the Kubernetes cluster
+  cmd: kinder
+  args:
+  - do
+  - kubeadm-join
+  - --name={{ .vars.clusterName }}
+  - --loglevel=debug
+  - --kubeadm-verbosity={{ .vars.kubeadmVerbosity }}
+  - --kubeadm-feature-gate="NodeLocalCRISocket=true"
```
neolit123, line 75: this is not needed.
we only need the --kubeadm-feature-gate during kubeadm-init

kinder/ci/tools/update-workflows/templates/workflows/instance-config-tasks.yaml
```diff
@@ -0,0 +1,322 @@
...
+    upgrades the cluster to Kubernetes "upgradeVersion"
+  cmd: kinder
+  args:
+  - do
+  - kubeadm-upgrade
+  - --upgrade-version={{ .vars.upgradeVersion }}
+  - --name={{ .vars.clusterName }}
+  - --loglevel=debug
+  - --kubeadm-verbosity={{ .vars.kubeadmVerbosity }}
+  - --kubeadm-feature-gate="NodeLocalCRISocket=true"
```
neolit123, line 145: also not needed. `kubeadm upgrade` downloads the kubeadm-config config map and construct a ClusterConfiguration from it. then all places in upgrade would know about the current list of enabled feature gates.

neolit123 (commented):
SGTM, minus you don't need to pass the FG on anything but init.

also we need to add the two missing things after we update k/k:
- on upgrade, remove --container-runtime-endpoint from kubeadm-flags.env check in this e2e check that it's not there anymore
- after upgrade if the FG is enabled make sure that the annotation is also removed from all nodes

HirazawaUi:
@neolit123  Do you think we can merge it now, or should we first fix the missing things in k/k?

HirazawaUi:
> > @neolit123 Do you think we can merge it now, or should we first fix the missing things in k/k?
> 
> the ci is green. please find time to fix the two remaining issues in k/k. we should try to get them in 1.33 and update the e2e.

I checked the test logs, and everything matches expectations.

Many thanks to @neolit123 for the help! I will address the remaining issues right away.

PR #3140, opened 2024-12-31
add support for kubeadm-upgrade --ignore-preflight-errors

pacoxu:
ignore SystemVerification when needed

- see https://github.com/kubernetes/kubernetes/pull/129401

### DIFF
--- a/kinder/ci/tools/update-workflows/templates/workflows/external-ca-tasks.yaml
+++ b/kinder/ci/tools/update-workflows/templates/workflows/external-ca-tasks.yaml
@@ -114,6 +114,7 @@ tasks:
       - --name={{ .vars.clusterName }}
       - --loglevel=debug
       - --kubeadm-verbosity={{ .vars.kubeadmVerbosity }}
+      - --ignore-preflight-errors=SystemVerification
     timeout: 15m
   - name: cluster-info-after
     description: |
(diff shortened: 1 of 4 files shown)


kinder/ci/tools/update-workflows/templates/workflows/external-ca-tasks.yaml
```diff
@@ -114,6 +114,7 @@ tasks:
       - --name={{ .vars.clusterName }}
       - --loglevel=debug
       - --kubeadm-verbosity={{ .vars.kubeadmVerbosity }}
+      - --ignore-preflight-errors=SystemVerification
```
SataQiu, line 117: shall we use variable here?  just like https://github.com/kubernetes/kubeadm/blob/cfc3ba886ce28a70ca263cda15a9b46286b8df4a/kinder/ci/workflows/skew-x-on-y-tasks.yaml#L19

SataQiu: Maybe that's okay, because it only appear once

pacoxu:
/close

> 
> to fix it you might want to PR kinder to:
> 
> don't pass the --ignore-preflight errors as flags for commands.
> pass them via config in the big config that has all the config 'kinds'.

PR #3141, opened 2024-12-31
move ignore preflight checks from flag to kubeadm configurations

pacoxu:
refer to https://github.com/kubernetes/kubernetes/pull/129401

After we add `SystemVerification` check in upgrade, we need to skip in some kinder testing.

This PR follows the suggestions https://github.com/kubernetes/kubernetes/pull/129401#pullrequestreview-2525461483 to move flags to configurations.

- don't pass the --ignore-preflight errors as flags for commands.
- pass them via config in the big config that has all the config 'kinds'.


neolit123 (commented):
thanks
/lgtm

Issue #2876, opened 2023-05-15
cleanup docs post GCS infra migration


neolit123:
see this and linked tracked issues:
- https://github.com/kubernetes/kubeadm/pull/2872
- https://github.com/kubernetes/kubeadm/pull/2872#issuecomment-1547866184
- [x] kubernetes/k8s.io#2396 (when this is closed we can update k/kubeadm)

the k/kubeadm docs contain some outdated examples. the PR above fixes some of them, but there is remaining work.

- [x] remove / change all gs:// entries in docs
- [x] remove /cluster/* script examples in docs
https://github.com/kubernetes/kubeadm/pull/2878

note we cannot change `gs://k8s-release-dev/ci/`
yet:
https://github.com/kubernetes/kubeadm/pull/2872#discussion_r1193581291

so this issue tracks future work.
but the /cluster change can be PRed now.

- https://github.com/kubernetes/kubeadm/pull/3148

chendave:
Just a record, here is still reference to gcs bucket, https://github.com/kubernetes/kubeadm/blob/3d02f6cd3ac64a9446bece239588291ec12cdad2/docs/testing-pre-releases.md?plain=1#L117-L119

Is there any alternative to replace it? https://console.cloud.google.com/storage/browser/kubernetes-release/release/


When we have any other options, we can update the link above.

chendave:
> remove / change all gs:// entries in docs

@neolit123  what do you mean for this todo item?  I searched in the code base, seems this has been addressed unless the `gs://k8s-release-dev/ci/` (I understood this will stay for an while).

neolit123:
> Is there any alternative to replace it?

@rjsadow do you know if we can list these labels with curl?

neolit123:
> > remove / change all gs:// entries in docs
> 
> @neolit123 what do you mean for this todo item? I searched in the code base, seems this has been addressed unless the `gs://k8s-release-dev/ci/` (I understood this will stay for an while).

we can't change it for now, afaik. it was discussed earlier as planned.

rjsadow:
I don't believe there's a direct 1-for-1 replacement with curl in this case since directory browsing isn't available through the proxy. It might be worth reaching out to sig-release and seeing if we could add an index file or some other mechanism into the release bucket that provides a list of all latest and stable versions.

rjsadow:
[...]

PR #3148, opened 2025-01-13
docs: don't reference gs://k8s-release-dev/ci/

neolit123:
https://storage.googleapis.com/k8s-release-dev/ci still exists and is used by kinder on each run to download CI artifacts.

If it gets removed we can update code and the related docs.
Until then don't reference the gs:// URLs for that bucket.

fixes https://github.com/kubernetes/kubeadm/issues/2876

- https://github.com/kubernetes/kubeadm/issues/2876

> https://github.com/kubernetes/k8s.io/issues/2396 (when this is closed we can update k/kubeadm)

- https://github.com/kubernetes/k8s.io/issues/2396

was closed. not clear if https://storage.googleapis.com/k8s-release-dev/ci will be replaced with something else, but we will know right away as it will break all jobs here:
https://testgrid.k8s.io/sig-cluster-lifecycle-kubeadm


pacoxu (commented):
/lgtm

PR #3168, opened 2025-03-03
kubeadm: fix periodic-kubernetes-e2e-manifest-lists failing test

SataQiu:
fix: #3167

PR #3170, opened 2025-03-03
print go version and update to go 1.24 of verify-manfiest-lists.sh

neolit123:
- the old go 1.16 is throwing a strange error
https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/periodic-kubernetes-e2e-manifest-lists/1896113622615068672

```
# [command-line-arguments]
./verify_manifest_lists.go:186:14: non-constant format string in call to fmt.Printf
FAIL	command-line-arguments [build failed]
FAIL
```

but the code is
`		fmt.Printf("%s%s", v.String(), sep)`

locally with newer go this error is not reproducible.

- bump to go 1.24 in varify_manifest_lists.sh and pin the api-machinery version to 0.32.
- print the go version in kinder / manifest test scripts.

### DIFF
--- a/kinder/ci/kinder-run.sh
+++ b/kinder/ci/kinder-run.sh
@@ -39,6 +39,7 @@ WORKFLOW_BUCKET_PATH="${SCRIPT_PATH}/workflows"
 # build kinder
 pushd "${ROOT_PATH}"
 echo "Building kinder..."
+go version
 GO111MODULE=on go build
 popd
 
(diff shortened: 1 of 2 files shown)


neolit123 (commented):
/cc @pacoxu

pacoxu (approved):
/lgtm

neolit123:
> * the old go 1.16 is throwing a strange error

looks like there was a fix here:
- https://github.com/kubernetes/kubeadm/pull/3168

pacoxu:
But this seems to be no harm.

PR #3188, opened 2025-04-28
chore: Replace codes with slices/maps package

sivchari:
/kind cleanup

Replace some codes with slices/maps package to reduce the management codes.


neolit123 (commented):
thanks
/approve

/cc @HirazawaUi ptal if you get a moment.

HirazawaUi (commented):
/lgtm

PR #3204, opened 2025-05-26
kinder: update version to v1.0.0

neolit123:
kinder has been used for kubeadm e2e tests for many releases, but its version remained 0.x. Update the version to v1.0.0 to identify its stability.


kinder/cmd/kinder/version/version.go
```diff
@@ -32,7 +32,7 @@ func NewCommand() *cobra.Command {
 		Short: "prints the kinder CLI version",
 		Long:  "prints the kinder CLI version",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			fmt.Printf("kinder version: %s\n", constants.KinderVersion)
+			fmt.Printf("v%s\n", constants.KinderVersion)
```
neolit123, line 35: just a cosmetic change, to give a simpler output for `kinder version`

PR #3213, opened 2025-06-18
Let kinder supports arm64

HirazawaUi:
kinder doesn't support arm64, which has been quite troublesome for me during testing because my laptop use the arm64 architecture. Making it compatible with arm64 now.

### DIFF
--- a/kinder/pkg/cri/nodes/containerd/alterhelper.go
+++ b/kinder/pkg/cri/nodes/containerd/alterhelper.go
@@ -21,6 +21,8 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


kinder/pkg/cri/nodes/containerd/alterhelper.go
```diff
@@ -138,13 +140,142 @@ func ImportImage(bc *bits.BuildContext, tar string) error {
...
+	if len(archLines) == 0 {
+		return fmt.Errorf("failed to get host architecture: empty output")
+	}
+
+	hostArch := strings.TrimSpace(archLines[0])
+	if hostArch == "" {
+		return fmt.Errorf("failed to get host architecture: empty architecture string")
+	}
+
+	// Map architecture names to standard names
```
neolit123, line 160: please use dot at the end of comment sentences everywhere or remove it everywhere

kinder/pkg/cri/nodes/containerd/alterhelper.go
```diff
@@ -138,13 +140,142 @@ func ImportImage(bc *bits.BuildContext, tar string) error {
...
+	var validTarFiles []string
+	for _, tarFilePath := range tarFilesLines {
+		tarFilePath = strings.TrimSpace(tarFilePath)
+		if tarFilePath != "" {
+			validTarFiles = append(validTarFiles, tarFilePath)
+		}
+	}
+
+	if len(validTarFiles) == 0 {
+		log.Infof("No .tar files found in %s\n", srcFolder)
```
neolit123, line 190: we can omit the \n in log comments here and everywhere because the logger auto-adds it?

kinder/pkg/cri/nodes/containerd/alterhelper.go
```diff
@@ -138,13 +140,142 @@ func ImportImage(bc *bits.BuildContext, tar string) error {
...
+		importCmd := fmt.Sprintf("ctr --namespace=k8s.io images import --all-platforms '%s'", tarFilePath)
+		importLines, err := bc.CombinedOutputLinesInContainer("bash", "-c", importCmd)
+		if err != nil {
+			errMsg := fmt.Sprintf("failed to import image from %s: %v", tarFilePath, err)
+			log.Errorf("Error: %s\n", errMsg)
+			importErrors = append(importErrors, errMsg)
+			continue // Continue with next file instead of failing completely
+		}
+
+		// Parse the imported image name from the output
```
neolit123, line 210: reading the image name from the tar seems less volatile to me.
the log output of ctr has no "API" guarantees and can change at any time.

neolit123 (commented):
+1, added a few comments.
unit tests might have been good but most of this is calling a shell command.
still, if you can break out some util functions like the image regexp matching and unit test it that would be better.

also, once this is ready to merge, please show before and after log output, so that we ensure all images are imported correctly.

importing the images saves a lot of bandwidth from k8s test infra and we should try to continue doing that as it saves $$.

HirazawaUi:
> +1, added a few comments. unit tests might have been good but most of this is calling a shell command. still, if you can break out some util functions like the image regexp matching and unit test it that would be better.
> 
> also, once this is ready to merge, please show before and after log output, so that we ensure all images are imported correctly.
> 
> importing the images saves a lot of bandwidth from k8s test infra and we should try to continue doing that as it saves $$.

Thanks for your comments. Since this matter isn't urgent, I'll address and resolve these comments later on (perhaps next weekend) :)

kinder/pkg/cri/nodes/containerd/alterhelper.go
HirazawaUi: fiex.

kinder/pkg/cri/nodes/containerd/alterhelper.go
HirazawaUi: Fixed.

neolit123 (commented):
/lgtm

thanks. seems like a good improvement.
please drop the hold when ready to merge it.

Issue #3218, opened 2025-07-02
kubeadm-kinder-instance-config-latest Test Job Fails Intermittently


HirazawaUi:
The kubeadm-kinder-instance-config-latest test job occasionally fails. 
Testgrid link: https://testgrid.k8s.io/sig-cluster-lifecycle-kubeadm#kubeadm-kinder-instance-config-latest

The errors primarily occur during cluster upgrades, with the following error log:
```
[preflight] Some fatal errors occurred:
	[ERROR ControlPlaneNodesReady]: there are NotReady control-planes in the cluster: [kinder-instance-config-control-plane-1]
[preflight] If you know what you are doing, you can make a check non-fatal with `--ignore-preflight-errors=...`
error: [upgrade/health] FATAL: preflight checks failed
k8s.io/kubernetes/cmd/kubeadm/app/cmd/upgrade.enforceRequirements
	k8s.io/kubernetes/cmd/kubeadm/app/cmd/upgrade/common.go:135
k8s.io/kubernetes/cmd/kubeadm/app/cmd/upgrade.runPlan
	k8s.io/kubernetes/cmd/kubeadm/app/cmd/upgrade/plan.go:104
k8s.io/kubernetes/cmd/kubeadm/app/cmd/upgrade.newCmdPlan.func1
	k8s.io/kubernetes/cmd/kubeadm/app/cmd/upgrade/plan.go:78
github.com/spf13/cobra.(*Command).execute
	github.com/spf13/cobra@v1.9.1/command.go:1015
github.com/spf13/cobra.(*Command).ExecuteC
	github.com/spf13/cobra@v1.9.1/command.go:1148
github.com/spf13/cobra.(*Command).Execute
	github.com/spf13/cobra@v1.9.1/command.go:1071
k8s.io/kubernetes/cmd/kubeadm/app.Run
	k8s.io/kubernetes/cmd/kubeadm/app/kubeadm.go:48
main.main
	k8s.io/kubernetes/cmd/kubeadm/kubeadm.go:25
runtime.main
	runtime/proc.go:283
runtime.goexit
	runtime/asm_amd64.s:1700
Error: failed to exec action kubeadm-upgrade: exit status 1
 exit status 1
```
ref: https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/ci-kubernetes-e2e-kubeadm-kinder-instance-config-latest/1940179256009035776

Opening this PR to document the investigation process for this issue.

[...]

PR #3219, opened 2025-07-02
After the upgrade is complete, wait for the node to be ready.

HirazawaUi:
This PR attempts to fix the flake issue with `kubeadm-kinder-instance-config-latest` test job. For a detailed analysis of the issue, please see: https://github.com/kubernetes/kubeadm/issues/3218#issuecomment-3027841437


neolit123 (commented):
thanks, makes sense to me.
let's monitor if this fixes the flake.

/lgtm

Issue #2271, opened 2020-08-31
make components on control-plane nodes point to the local API server endpoint


neolit123:
in CAPI immutable upgrades we saw a problem where a 1.19 joining node cannot bootstrap, if a 1.19 KCM takes leadership and tries to send a CSR to a 1.18 API server on an existing Node. this happens because in 1.19 the CSR API graduated to v1 and a KCM is supposed to talk to a N or N+1 API server only.

a better explanation here:
https://kubernetes.slack.com/archives/C8TSNPY4T/p1598907959059100?thread_ts=1598899864.038100&cid=C8TSNPY4T

- [x] we should make the controller-manager.conf and scheduler.conf that kubeadm generates talk to the local API server and not to the controlPlaneEndpoint (CPE, e.g. LB).
PR for 1.20: https://github.com/kubernetes/kubernetes/pull/94398
PR for 1.19: https://github.com/kubernetes/kubernetes/pull/94442
- [x] relax the server URL validation in kubeconfig files:
https://github.com/kubernetes/kubeadm/issues/2271#issuecomment-690822335
https://github.com/kubernetes/kubernetes/pull/94816

----

optionally we should see if we can make the kubelet on control-plane Nodes bootstrap via the local API server instead of using the CPE. this might be a bit tricky and needs investigation. we could at least post-fix the kubelet.conf to point to the local API server after the bootstrap has finished.
see https://github.com/kubernetes/kubernetes/issues/80774 for a related discussion

this change requires a more detailed plan, a feature gate and a KEP

list of PRs tracked here:
- https://github.com/kubernetes/enhancements/issues/4471

1.31 alpha
- https://github.com/kubernetes/kubernetes/pull/125582
 
1.33 beta 
- https://github.com/kubernetes/kubernetes/pull/129956

1.35 GA
- https://github.com/kubernetes/kubernetes/pull/134106

1.36 post-GA cleanup - TODO

neolit123:
PR is here: https://github.com/kubernetes/kubernetes/pull/94398

neolit123:
we spoke about the kubelet.conf in the office hours today:
- Pointing the kubelet to the local api server should work, but the kubelet-start phase has to happen after the control-plane manifests are written on disk for CP nodes.
- Requires phase reorder and we are considering using a feature gate.
- This avoids skew problems of a new kubelet trying to bootstrap against an old api-server.
- One less component to point to the CPE.

[...]

PR #3239, opened 2025-09-17
kinder: remove ControlPlaneKubeletLocalMode workflows after it goes GA

neolit123:
once the FG goes GA and is locked to true, we no longer need the case where we test it being false.

xref 
- https://github.com/kubernetes/kubeadm/issues/2271

/hold for
- https://github.com/kubernetes/kubernetes/pull/134106


HirazawaUi (commented):
/lgtm

PR #3245, opened 2025-10-03
kinder: bump to kindest/base:v20250521-31a79fd4

neolit123:
Newer versions have containerd 2.x, and we can try using this newer version so that we can test features like HostUsers=true.

/hold


HirazawaUi (commented):
/lgtm

PR #3275, opened 2026-01-06
Update kube-vip alias and notes in documentation

### LINKED ISSUE #3274: docs: update kube-vip instructions for containerd and k8s v1.29+
I was following the High Availability considerations documentation to set up a cluster using `kube-vip` as a static pod. I encountered two issues with the current instructions:

1.  **Containerd Alias Failure:** The provided `alias` command for `ctr` fails because it attempts to run an image that hasn't been pulled yet. `ctr` does not automatically pull images like Docker does.
2.  **Kubernetes v1.29+ Bootstrap Failure:** On Kubernetes v1.29 and later, the `kubeadm init` process times out because the `kube-vip` static pod crashes. It attempts to use `admin.conf`, which lacks the necessary permissions during the initial bootstrap phase due to changes introduced in v1.29.

Image

**What you expected to happen**:
The documentation commands should work copy-paste for setting up HA with `kube-vip` on supported Kubernetes versions.

**How to reproduce it**:
-  Attempt to run the `ctr` alias command on a fresh node without pulling the image first.
-  Attempt to bootstrap a v1.29+ cluster using the current static pod manifest that points to `/etc/kubernetes/admin.conf` immediately.

**Proposed Solution**:
I have successfully verified a fix in my environment and would like to update `docs/ha-considerations.md` with the following changes:

-  Update the `ctr` instructions to explicitly pull the image before creating the alias.
-  Add a note regarding the v1.29+ breaking change, documenting the workaround of using `super-admin.conf` for the initial bootstrap and switching back to `admin.conf` post-initialization (as referenced in `kube-vip` issue #684).

This workaround is referenced and validated in the official [Kube-VIP docs](https://kube-vip.io/docs/installation/static/)

I have tested these updated commands on a 3-node HA cluster running Kubernetes v1.35 and confirmed they resolve the bootstrap timeout issues. I am ready to submit a PR with these updates.

Image

ariefshaik7:
Updated the kube-vip alias command for containerd to include an image pull step before running the container. Added note about changes in kubeadm bootstrap process since Kubernetes v1.29 based on the Official Kube-vip docs.

Closes #3274

### DIFF
--- a/docs/ha-considerations.md
+++ b/docs/ha-considerations.md
@@ -287,7 +287,7 @@ The easiest method to generate a manifest is using the container itself, below w
 
 #### containerd
 
-`alias kube-vip="ctr run --rm --net-host ghcr.io/kube-vip/kube-vip:$KVVERSION vip /kube-vip"`
+`alias kube-vip="ctr image pull ghcr.io/kube-vip/kube-vip:$KVVERSION; ctr run --rm --net-host ghcr.io/kube-vip/kube-vip:$KVVERSION vip /kube-vip"`
 
 #### Docker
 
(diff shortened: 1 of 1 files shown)


docs/ha-considerations.md
```diff
@@ -299,14 +299,18 @@ This configuration will create a manifest that starts `kube-vip` providing **con
 
 `export INTERFACE=eth0`
 
+> **Note**: When running these commands on a to-be control plane node, `sudo` access may be required along with pre-creation of the `/etc/kubernetes/manifests/` directory.
```
neolit123:
```suggestion
**NOTE** When running these commands on a to-be control plane node, `sudo` access may be required along with pre-creation of the `/etc/kubernetes/manifests/` directory.
```

docs/ha-considerations.md
```diff
@@ -299,14 +299,18 @@ This configuration will create a manifest that starts `kube-vip` providing **con
...
+
 ```
 kube-vip manifest pod \
     --interface $INTERFACE \
-    --vip $VIP \
+    --address $VIP \
     --controlplane \
+    --services \
```
neolit123: this change is not explained in the ticket.
why is it needed?

PR #3292, opened 2026-03-13
bump all kinder base image to v20260214-ea8e5717

pacoxu:
- currently, all CI are using v20250521-31a79fd4 which is containerd 2.1.1.

This is not an urgent PR and may needs testing.


neolit123:
the coredns failure might be a flake

